Issue #87, opened 2014-07-17
Routine generator


rolandgeider:
There should be a possibility to automatically generate weightlifting routines based on the user input. Liftwhat is a nice implementation and should be used for inspiration:
- https://github.com/scottt2/liftwhat
- http://www.liftwhat.com

PR #88, opened 2014-07-19
Automatic routine generator

rolandgeider:
Add an automatic (weightlifting) routine generator.

See #87


roshkins:
Does this still need working on? I love this idea and I'd love to integrate some of the exercise science I've been reading up on to generate routines.

PR #147, opened 2015-06-04
Adds latest weight details to overview

max111:


rolandgeider:
Ah, I see, those are coding style errors. "pep8" is more or less the official coding style for python projects with the idea that when you switch between projects, the code layout remains the same. There is a tool, also called pep8, that automatically checks the code. You can install it in your virtual environment (``pip install pep8``) and then just do ``pep8 wger``. The errors should be self explaining and once you have corrected a couple, you'll format the code automatically.

rolandgeider:
And another thing, don't forget to add yourself to AUTHORS.rst :)

wger/weight/templates/weight_overview.html
```diff
@@ -38,6 +38,24 @@
 <div id="weight_diagram"></div>
 
 {% if is_owner %}
+{% if last_five_weight_entries_details %} 
+<table border="1" style="width:100%">
```
rolandgeider: better do here ``<table class="table">``, we can use the [bootstrap](http://getbootstrap.com/) CSS classes

wger/weight/templates/weight_overview.html
```diff
@@ -38,6 +38,24 @@
...
+    <tr>
+        <th>Date</th>
```
rolandgeider: Better write ``{% trans "Date" %}``, then the header can be translated when using the site on another language, same with change, etc.

wger/weight/views.py
```diff
@@ -150,9 +150,29 @@ def overview(request, username=None):
...
+            weight_diff = curr_entry.weight - prev_entry.weight 
+            if weight_diff > 0:
+                weight_diff = '+'+str(weight_diff)
+            day_diff = (curr_entry.date - prev_entry.date).days
+            last_five_weight_entries_details.append((curr_entry, weight_diff, day_diff))
+            i += 1
+
+        curr_entry = last_five_weight_entries[i] 
+        last_five_weight_entries_details.append((curr_entry,'-/-', '-/-'))
+
```
rolandgeider: While the code works, the loop doesn't feel *pythonic*. I would suggest doing something along these lines (not tested!):

```python
curr_entry = None
for entry in last_five_weight_entries:
    prev_entry = curr_entry
    curr_entry = entry
    if prev_entry and curr_entry:
        weight_diff = curr_entry.weight - prev_entry.weight
        day_diff = (curr_entry.date - prev_entry.date).days
        # and so on...
```

wger/weight/views.py
```diff
@@ -150,9 +150,29 @@ def overview(request, username=None):
...
+    if last_five_weight_entries:
+        i = 0 
+
+        while (i != len(last_five_weight_entries)-1):
+            curr_entry = last_five_weight_entries[i]
+            prev_entry = last_five_weight_entries[i+1]
+            weight_diff = curr_entry.weight - prev_entry.weight 
+            if weight_diff > 0:
+                weight_diff = '+'+str(weight_diff)
```
rolandgeider: I would move this check to the template, seems more natural to me to have it there

wger/weight/templates/weight_overview.html
max111: Awesome, incorporating it now.

wger/weight/templates/weight_overview.html
max111: yes gotcha!

wger/weight/views.py
max111: On doing the check in the template, '-/- > 0 would also cause the '+' sign. I'm thinking that we would need a helper function to check whether weight_diff is numeric first, then call that first in template. Where shall we put that helper function?

max111: I just learned about template filter actually, maybe I should look into built in django filters for that or write up a custom filter. What do you think?

rolandgeider: Actually, we don't need a helper function at all, in the view we can just do ``weight_diff = day_diff = None``. Then in the template we can check if the value is set, something like this

```
{% if entry_detail.1 %}
    {% if entry_detail.1 > 0 %}
        +
    {% endif %}
    {{ entry_detail.1 }}
{% else %}
    -/-
{% endif %}
```

This also makes is a bit clearer in the future to know where the "special" values + and -/- come from, since it is all done in the template

max111: Ok that's a good idea. I'm making changes and pushed by mistake. Please
ignore the current pull request for now.

Best regards,
Yu Yu Aung

max111: It's all fixed. I was looking into weight instead of weight_diff for the
"+" sign.

rolandgeider:
Looks good!

Issue #124, opened 2015-01-23
Allow external access to schedules


rolandgeider:
Logic basically the same as with workouts, if the user has allowed external access, anonymous users should access schedules via a shared link.

rolandgeider:
Some more info: Some user pages such as workouts are accessible for external (non-logged in users) if the owner has allowed this in the preferences and has shared the link e.g. via facebook or something.

Individual schedules should be made accessible as well, the logic will be basically the same as with workouts (manager/views/workout.py, view-method) and can be probably just copy/pasted. Also a todo: making sure the different edit/buttons in the template are hidden for guests (they can't edit or delete anything anyway, but still should not see them)

rolandgeider:
Closed with #148

PR #148, opened 2015-06-07
Allow read only access for schedules

rlaszlo:
Removed the login_required decorator from the schedule view
and check for the read only access settings.

The edit buttons and the content visible only for the logged in
users is also wrapped in conditions.

Added the open graph meta tags like in the workout view.

### DIFF
--- a/wger/manager/views/schedule.py
+++ b/wger/manager/views/schedule.py
@@ -64,16 +64,20 @@ def overview(request):
     return render(request, 'schedule/overview.html', template_data)
 
 
-@login_required
 def view(request, pk):
     '''
     Show the workout schedule
     '''
     template_data = {}
-    user = request.user
+    schedule = get_object_or_404(Schedule, pk=pk)
+    user = schedule.user
+    is_owner = request.user == user
+
+    if not is_owner and not user.userprofile.ro_access:
+        return HttpResponseForbidden()
+
     uid, token = make_token(user)
 
-    schedule = get_object_or_404(Schedule, pk=pk, user=user)
     template_data['schedule'] = schedule
     if schedule.is_active:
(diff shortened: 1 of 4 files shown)


rlaszlo:
Hi Roland,

I created this pull request for the issue #124.
I'm quiet new to open source so if you see any problems with the request then please tell.

I hid the "Adding workouts" section too from the template on anon access, because it does not makes any sense for users not logged in.

wger/manager/views/schedule.py
```diff
@@ -84,6 +88,8 @@ def view(request, pk):
 
     template_data['uid'] = uid
     template_data['token'] = token
+    template_data['is_owner'] = is_owner
+    template_data['owner_user'] = user
```
rolandgeider: A little thing, you need to add ``template_data['show_shariff'] = is_owner`` so the sharing buttons are rendered in the template

rolandgeider:
Hi @rlaszlo ! Thanks, this looks nice, I made a small comment, but then it looks that we can merge this.

rolandgeider:
I always forget, add yourself to AUTHORS.rst as well :)

wger/manager/views/schedule.py
rlaszlo: Thanks, didn't notice it in the workout. Fixing in the next commit.

rlaszlo:
Thank you!

I forgot to check the tests, so the build failed. Fixed it now.

rlaszlo:
Hi Roland,

the AUTHORS.rst is in conflict, so github can resolve it. Should I merge your master branch and push a mergable request?

rolandgeider:
Hi @rlaszlo I just merged this, the conflict in the authors file was trivial so we would be spending more time talking about it than doing it :)

Issue #127, opened 2015-01-29
Allow copying workout of different user


rolandgeider:
When accessing a workout of another user (and one is logged in) allow to copy the workout for own use

max111:
How can I access the workout of another use (one who's not logged on)?

rolandgeider:
The user first has to allow external access to his/her data, that's just a flag in the preferences. Then you can access the workout if you know the URL, in reality you would arrive there because it was shared in facebook or similar

PR #158, opened 2015-07-03
Add ability to copy public workouts

jamessimas:
If a user has made this workouts public, this change gives other users
the ability to copy the public workout into their own workouts.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -10,6 +10,7 @@ Developers
 * Matheus Zorzete Marchiore - https://github.com/marchiore
 * Yu Yu Aung - https://github.com/max111
 * Laszlo Ratsko - https://github.com/rlaszlo
+* James Simas - https://github.com/jamessimas
 
 Translators
 -----------
--- a/wger/manager/templates/workout/view.html
+++ b/wger/manager/templates/workout/view.html
@@ -171,7 +171,6 @@ $(document).ready(function() {
 </a>
 </p>
 
-{% if is_owner %}
 <p>
 <a href="{% url 'manager:workout:copy' workout.id %}"
    class="wger-modal-dialog">
@@ -184,6 +183,7 @@ $(document).ready(function() {
 </a>
 </p>
 
+{% if is_owner %}
 <p>
 <a href="{% url 'manager:workout:delete' workout.id %}"
    class="wger-modal-dialog">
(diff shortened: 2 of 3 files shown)


jamessimas:
This is a fix for issue #127.

rolandgeider:
@jamessimas thanks! There is still a test failing, after you fix that I can merge this.

Also don't forget to add yourself to the AUTHORS file :)

jamessimas:
Hi @rolandgeider ,

The reason the test is failing is because of the following test in test_copy_workout.py:

        response = self.client.get(reverse('manager:workout:copy', kwargs={'pk': '3'}))
        if not owner:
            self.assertEqual(response.status_code, 404)

The test assumes that if you're not the owner of a workout, you cannot copy it. With the changes I am proposing be committed, this is no longer valid. The logic should be "if the user is not the owner **and** the workout is not public, then assert a 404 is returned".

Since I'm new to push requests, I'm not sure if I can try to update your test logic on Travis CI or not. (My expectation is that I cannot.)

How should I proceed?

Thanks for the opportunity to help out!

James

rolandgeider:
Hi James!

No problem, this started as a weekend project where I wanted to "do things right", with tests, code pep8 compliant, etc. It's nice to see that it has become something that interests other people :)

Since we are changing the logic, we need to update the tests. I would think that it would be enough to remove all the ``if not owner``, since we know that for workout 3 this all works. Then add a new check to see that the new workout, does not belong to the original owner but to the user that copied it. And finally there is a new, smaller test where we just test for the status code if there isn't read access.  Something like this (not tested!!):

```python
def test_copy_shared():
    profile = Userprofile.objects.get(user_username='test')
    profile.ro_access = False
    profile.save()
    self.user_login('admin')
    response = self.client.get(reverse('manager:workout:copy', kwargs={'pk': '3'}))
    self.assertEqual(response.status_code, 404)
```

jamessimas:
Hi Roland,

Thanks for the response!

I'll start working on this later today. One question, though... If I update the test logic and include that in the pull request, will Travis CI use the test logic that I modify?

James

rolandgeider:
Yes, travis will pick up the changes you make

jamessimas:
Hey,

I've updated the unit tests and confirmed that my dev branch passes the Travis CI tests.

I had to create a new pull request because I had to re-fork your repo. The new pull request is here: https://github.com/rolandgeider/wger/pull/161

Feel free to close/reject this pull request and we can begin working in the other one.

Thanks again!

James

PR #161, opened 2015-07-04
Fix issue #127

jamessimas:
This fixes issue #127. Users can copy the workouts of other users who have made their workouts public.


rolandgeider:
Looking good! Also, very explicit commit messages :+1: 

(The other pull request is shown as already merged because those commits went in when fixing that typo)

jamessimas:
Thanks for all your help on this one!

rolandgeider:
Well, actually, I didn't do that much ;)

PR #162, opened 2015-07-04
Drf 3 upgrade

rlaszlo:
Hi @rolandgeider ,

I'm finished with the upgrade. If you have time then please review the pull request.

There a lot of changes:

1. Added a serializer and a queryset where it was missing, and removed the model attributes (model are attributes are removed from 3X).

2. The link decorator is renamed to detail_route.

3. pre_save methods are not available, there are perform_create and perform_update methods, but these are working with values instead of objects. Because of this I left a TODO at the ExerciseViewSet and ExerciseImageViewSet perform_create. You can't call a model method before saving to the db. 
I'm calling the set_author after the save now, but it does not seem to a good solution. 
Do you have any idea for this?

4. Because the IngredientViewSet get_values not using the Serializer I needed to coerce the float values manually to string. 

5. In some models the foreign keys are not editable (MealItem), because of this the ModelSerializer sets these fields readonly. I needed to add these fields manually for the model serializers to enable setting the values from the API.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ sphinx
 django-sortedm2m
 
 # REST API
-djangorestframework==3.1.3
+djangorestframework>=3.1,<3.2
 django-filter
 django-tastypie
 django-cors-headers
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ sphinx
 django-sortedm2m
 
 # REST API
-djangorestframework==2.3.13
+djangorestframework==3.1.3
 django-filter
 django-tastypie
 django-cors-headers
--- a/wger/exercises/api/views.py
+++ b/wger/exercises/api/views.py
@@ -27,6 +27,8 @@ from easy_thumbnails.files import get_thumbnailer
 from django.utils.translation import ugettext as _
 
 from wger.config.models import LanguageConfig
+from wger.exercises.api.serializers import MuscleSerializer, ExerciseSerializer, ExerciseImageSerializer, \
+    ExerciseCategorySerializer, EquipmentSerializer, ExerciseCommentSerializer
 from wger.exercises.models import Exercise
 from wger.exercises.models import Equipment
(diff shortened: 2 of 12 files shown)


rolandgeider:
@rlaszlo Wow, that looks like a lot of work! I'll try to review it this weekend. Thanks!!!

requirements.txt
```diff
@@ -18,7 +18,7 @@ sphinx
 django-sortedm2m
 
 # REST API
-djangorestframework==2.3.13
+djangorestframework==3.1.3
```
rolandgeider: I don't like pinning to specific versions unless I know there are problems, I would suggest setting this to ``djangorestframework>=3.1,<3.2`` then we always get the latest in the 3.1.x branch, which in most cases just means bugfixes

wger/exercises/api/views.py
```diff
@@ -57,12 +60,15 @@ class ExerciseViewSet(viewsets.ModelViewSet):
...
-    def pre_save(self, obj):
+    def perform_create(self, serializer):
         '''
-        Set language, author and status
+        Set author and status
         '''
-        obj.language = load_language()
+        language = load_language()
+        obj = serializer.save(language=language)
+        # Todo is it right to call set author after save?
```
rolandgeider: Yes, it should not be a problem. These only set the status and send an email. There are probably better ways to solve this, for example using signals, but then the code has no access to the request, or at least it wasn't the last time I looked at this.

requirements.txt
rlaszlo: Fixed the requirements. Thanks!

wger/exercises/api/views.py
rlaszlo: Yes, I tried to place the logic to other places (model, serializer), but had no luck because of the request. I will try to look into it, maybe there is a better way.

rolandgeider:
Merged! :)

rlaszlo:
Thanks! :)

Issue #159, opened 2015-07-03
Default author for exercises should be hostname


rolandgeider:
At the moment, the author used for exercises when added by an administrator is the hard coded string 'wger.de'. This should be changed to the current hostname so that local installations get sensible data.

jamessimas:
Looking into this.

jamessimas:
I've just learned what interactive rebasing is, so my future pull requests should be cleaner. Huzzah!

James

rolandgeider:
I know that some projects practically demand to rebase everything for push requests, but I am more relaxed, if you needed 5 commits, then I get those 5 commits, they don't hurt. But obviously, if you want to tidy up the commits, feel free! :smile: (BTW, you can also rebase commits in a pull request after you have opened it, even multiple times)

jamessimas:
Good to know. Thanks.!

rolandgeider:
Implemented with pull request #166

PR #166, opened 2015-07-05
Change default author of exercise to hostname

jamessimas:
Fix for issue #159.

Commit c687301 fixes issue #159 by making the default author for new
exercises the hostname of the local computer running wger.

This commit (b00efde) updates the test to ensure this is what we test for.

### DIFF
--- a/wger/exercises/models.py
+++ b/wger/exercises/models.py
@@ -331,7 +331,8 @@ class Exercise(AbstractSubmissionModel, AbstractLicenseModel, models.Model):
         if request.user.has_perm('exercises.add_exercise'):
             self.status = self.STATUS_ACCEPTED
             if not self.license_author:
-                #  Get everything up to semicolon, exclusive. Example: my-gym.com:9090 returns my-gym.com
+                #  Get everything up to semicolon, exclusive.
+                #  Example: my-gym.com:9090 returns my-gym.com
                 results = re.search('^([^:]+).*', request.get_host())
 
                 if results:
--- a/wger/exercises/models.py
+++ b/wger/exercises/models.py
@@ -19,7 +19,6 @@ import six
 import uuid
 import logging
 import bleach
-import re
 
 from django.db import models
 from django.template.loader import render_to_string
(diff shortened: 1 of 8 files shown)


rolandgeider:
I probably wrote the ticket unclearly! I meant to write that the name of the (webserver)host be used, so if someone has an instance running on "my-local-gym.com", that value is used. The socket library reads the local name of the server running the code. I have googled a bit and found that django's request object has a [get_host()](https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.HttpRequest.get_host) that returns what we need

jamessimas:
Ok, thanks for the clarification!

I'll begin working on fixing this tonight.

jamessimas:
Ok. Done. 

Thoughts?

wger/exercises/models.py
```diff
@@ -330,8 +331,16 @@ def set_author(self, request):
...
             if not self.license_author:
-                self.license_author = 'wger.de'
+                #  Get everything up to semicolon, exclusive.
+                #  Example: my-gym.com:9090 returns my-gym.com
+                results = re.search('^([^:]+).*', request.get_host())
+
+                if results:
+                    name = results.group(1)
+                else:
+                    name = request.get_host()
```
rolandgeider: Regexes are a bit of overkill here, we can easily extract the part from the hostname that we need with a split: ``get_host().split(':')[0]``

jamessimas:
You're totally right. I'll fix this tonight. 

James

jamessimas:
I've updated my pull request with the change you suggested.

Any idea why it's failing in your Travis CI instance, but not in mine? (https://travis-ci.org/jamessimas/wger/builds/69990038)

I'm heading out of the country for the next few weeks. Internet access is pretty spotty, so I won't get a chance to continue working on this until I return.

rolandgeider:
I am also sometimes getting these errors, it seems it started some days ago. Perhaps it's something on travis' side? In any case, I'll merge this now. 

If you are going on vacation, relax, and don't think about issues! ;)

rolandgeider:
@jamessimas How are you doing? Back from vacation?

jamessimas:
@rolandgeider 

Thanks for checking in. I'm in the process of moving right now so I've been busy. I'll check back as soon as I wrap that up and have more time.

rolandgeider:
Cool, no problem!

PR #167, opened 2015-07-06
Fix inaccurate wording

jamessimas:
Previous wording indicated that email was only used for reseting
passwords, but the app also supports email reminders.

Updated wording takes into account these two uses for email addresses.

### DIFF
--- a/wger/core/forms.py
+++ b/wger/core/forms.py
@@ -42,7 +42,7 @@ class UserPreferencesForm(forms.ModelForm):
 
 class UserEmailForm(forms.ModelForm):
     email = EmailField(label=_("Email"),
-                       help_text=_("Only needed to reset your password in case you forget it."),
+                       help_text=_("Used for password resets and, optionally, workout reminders."),
                        required=False)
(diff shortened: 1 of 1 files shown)


wger/core/forms.py
```diff
@@ -42,7 +42,7 @@ class Meta:
 
 class UserEmailForm(forms.ModelForm):
     email = EmailField(label=_("Email"),
-                       help_text=_("Only needed to reset your password in case you forget it."),
+                       help_text=_("Used for password resets and, optionally, workout reminders."),
```
rolandgeider: Actually, there are other types of emails besides workout reminders, so I'd use something more general. Perhaps something like this? *"Used for password resets and, optionally, email reminders."*

rolandgeider:
Thanks, that text is very old, from when there were no reminders.

jamessimas:
I've updated the wording to what you suggest. Let me know if you need anything else changed!

rolandgeider:
Looks fine!

PR #170, opened 2015-07-13
Add height as a DecimalField to BmiForm

### LINKED ISSUE #146: BMI And Calorie Calculator Not Working
Using this software in Linux Mint 13.
When I enter my data into either the BMI calculator or the calorie estimator nothing happens.
I have entered my height in cm and my weight in kgs.
The BMI calculator says my BMI = 0.
I'd be happy with 10.

rlaszlo:
If the form got a floating point value in the height field, then
the validation failed. Because of this the bmi calculation didn't
return anything. Fixes #146

### DIFF
--- a/wger/nutrition/forms.py
+++ b/wger/nutrition/forms.py
@@ -55,6 +55,8 @@ class UnitChooserForm(forms.Form):
 
 
 class BmiForm(forms.ModelForm):
+    height = forms.DecimalField(widget=Html5NumberInput(),
+                                max_value=999)
     weight = forms.DecimalField(widget=Html5NumberInput(),
                                 max_value=999)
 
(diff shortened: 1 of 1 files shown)


wger/nutrition/forms.py
```diff
@@ -55,6 +55,8 @@ def __init__(self, *args, **kwargs):
 
 
 class BmiForm(forms.ModelForm):
+    height = forms.DecimalField(widget=Html5NumberInput(),
+                                max_value=999)
```
rolandgeider: Add here a ``label=_('Height (cm)')``, then the label is translated as well (that's the disadvantage of re-defining model fields, you have to duplicate a lot of things)

rlaszlo: Should I put label=_('Weight') to the weight field too?

rolandgeider: No, that one works

Issue #165, opened 2015-07-04
Don't allow empty sets


rolandgeider:
At the moment there is no validation when adding a new exercise to the workout without any repetitions.

When working on this issue it's probably a good idea to refactor the logic for this and let django do more of the heavy lifting.

Yevs:
Hi there. I seem to be new one here but I think I can fix this bug.

rolandgeider:
Awesome! Sure, go ahead!

PS: I will probably have no internet this weekend

Yevs:
oops, wrong commit, so sorry. let me fix this real quick

PR #175, opened 2015-07-31
Fixed 165

Yevs:
Fixed bug #165. I am not sure if should there be a test for it and db migration for "blank=False".

### DIFF
--- a/wger/manager/forms.py
+++ b/wger/manager/forms.py
@@ -112,6 +112,12 @@ class SettingForm(ModelForm):
         model = Setting
         exclude = ('set', 'exercise', 'order', 'comment')
 
+    def is_valid(self):
+        res = super(SettingForm, self).is_valid()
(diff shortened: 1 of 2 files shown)


rolandgeider:
Thanks for the pull request! 

I don't think that we need explicit tests for this.

The problem with this solution is that it forces you to fill in all the fields in the form. The logic behind the repetitions is "smart" so that it is not necessary to add the same repetitions if they are all the same (so you can enter for 4 sets just a 10 to get 4x10). I'm not sure if formsets allow to specify that *at least* one row must have data. If the formsets can't do that, then it has to be done manually and the code in the view is really horrible with lists and dictionaries with forms in them (it was one of the first things I wrote).

rolandgeider:
Oh, and don't forget to add yourself to AUTHORS.rst

Issue #126, opened 2015-01-29
Use bower to download the JS libraries


rolandgeider:
At the moment, each JS library needs to be downloaded and updated individually. Use bower to automate this.

tranbenny:
Hi! I just submitted a pull request attempting to implement this!

rolandgeider:
Implemented with #176

PR #176, opened 2015-08-04
Automate js libs

tranbenny:
Attempted to work on Issue #126 

Steps done: 
-added django-bower to requirements 
-configured django-bower to install bower packages in wger/core/static/bower_components
-deleted external library files that could be installed via bower
-added "python manage.py bower install" to README installation instructions
-edited routes to bower_components folder in template.html, template_features.html


rolandgeider:
Hey @tranbenny!

Thanks for the pull request, you really cleaned up the JS files! I'll see that I test this the next days.

rolandgeider:
Oh, and don't forget to add yourself to AUTHORS.rst :)

rolandgeider:
@tranbenny, I just took a look at this, from what I did test, everything still works with the installed libraries :) However, to make this actually work, I had to

* change in bowerrc the directory to an empty string: ``"directory" : ""`` 
* and in settings_global.py  ``BOWER_COMPONENTS_ROOT`` to  = ``os.path.join(BASE_DIR, 'wger', 'core', 'static', 'bower_components')``

Is this something with my setup or did you perhaps commit some older version of the files? In any case, these are just details, the main thing is that finally I won't have to update the libraries manually.

tranbenny:
@rolandgeider I'm glad it worked! I tried running it with your changes, and the change in the bower_components_root worked fine. But whenever I reinstalled the bower packages, the packages ended up in the wger directory. I'm not 100% sure on the details, but I think you need to specify where you want it to be installed in the .bowerrc file. Otherwise on first installations from other users, the packages may not be installed in the right place, and it won't be found on the links in the html files.

rolandgeider:
I just noticed that the new jquery ui broke some api, I'll have to update the code on my side.

rolandgeider:
I have expanded the docs somewhat and just merged this. I'm usually faster with pull requests, but it took a bit longer. Thanks again :)

Issue #202, opened 2015-09-26
Move pdf.py to views


rolandgeider:
The pdf.py file in the manager app actually belongs into the views folder

rolandgeider:
Closed with #209

PR #209, opened 2015-10-04
Move pdf.py to views #202

warchildmd:


rolandgeider:
Nothing to say here :) merged!

Issue #205, opened 2015-10-02
Minor visual issues


romansp:
Hi, @rolandgeider. Prepared list of minor visual issues. Can you review those, and I will later submit PR?

1. Modal dialog 'Share this page' via social media looks kinda wierd. Thinks it needs some rework.
![image](https://cloud.githubusercontent.com/assets/3474842/10254432/3fdc69d8-694c-11e5-81bc-c9e2b93d4ddf.png)

2. 'Get in touch' link in header gets color of background on mouse hover.
![image](https://cloud.githubusercontent.com/assets/3474842/10254455/7acf07ee-694c-11e5-836b-d8cbb80c93ed.png)

3. Navbar action links disappear on small screen if screen is shrinked.
![image](https://cloud.githubusercontent.com/assets/3474842/10254464/93f2f492-694c-11e5-903a-39768dda1bba.png)

4. Consider cancelling margin-top for the first h2 in mid column, so it will be 'in line' with left and right columns.![image](https://cloud.githubusercontent.com/assets/3474842/10255535/b207daae-6953-11e5-93c3-ffc113b961af.png)

5. Set line-height: 1 for those orange buttons, so text will appear in the middle;
![image](https://cloud.githubusercontent.com/assets/3474842/10255623/50dc1fe6-6954-11e5-822c-bb76aa61a7ac.png)

6. Insert space or line-break before [active] label.
![image](https://cloud.githubusercontent.com/assets/3474842/10255697/bad4b91c-6954-11e5-8bef-f7ec007288be.png)

7. The same as previous.
![image](https://cloud.githubusercontent.com/assets/3474842/10255741/13e672ac-6955-11e5-96a7-a3697c5c9607.png)

Will look for some more.

romansp:
Oh, you've just commited font for the missing icons in modal :)

rolandgeider:
nice to see you again @romansp! nice catches!

1. hehe, yeah

2. yes you are right, I could swear I had fixed this, but perhaps I just dreamed it :)

3. this is because the application is not really responsive, for some pages you get a different template if you use a mobile agent. Really fixing this is one thing I want to do in the future but till then, we should fix this

4. definitely

5. this would be the fastest way. The month table has other problems, e.g. when the screen is small (smartphone) and there are many entries in one week, the margins (or paddings?) make the whole table too wide.

6. and 7. yes

rolandgeider:
Closing issue, this was fixed in #205

Issue #206, opened 2015-10-02
Activate localization for share buttons


rolandgeider:
The shariff widget to share the pages (``sharif_modal.html``) defaults to German. Fixing this is basically just setting the ``data-lang`` attribute (see [documentation](https://github.com/heiseonline/shariff)), but only if the current language is supported by shariff

PR #213, opened 2015-10-25
Fixes to minor visual issues from #205

romansp:
This PR fixes some of the issues described in #205.

### DIFF
--- a/wger/manager/templates/schedule/overview.html
+++ b/wger/manager/templates/schedule/overview.html
@@ -23,7 +23,7 @@
             <a href="{% url 'manager:schedule:view' schedule.id %}">{{ schedule.name }}</a>
 
             {% if schedule.is_active %}
-            <span class="label label-primary">{% trans "active" %}</span>
+                &nbsp;<span class="label label-primary">{% trans "active" %}</span>
             {% endif  %}
         </td>
         <td>
--- a/wger/core/static/css/workout-manager-common.css
+++ b/wger/core/static/css/workout-manager-common.css
@@ -306,6 +306,7 @@ table.month th.month{
 
 table.month a.calendar-link{
     height:100%;
+    line-height: 1;
 }
 
 table.month td.session-bad {
--- a/wger/core/static/css/workout-manager.css
+++ b/wger/core/static/css/workout-manager.css
@@ -175,7 +175,9 @@ header h1{
     font-weight: 300;
 }
 
-header a{
+header a,
+header a:hover, 
+header a:focus{
     color:white;
     font-size:90%;
 }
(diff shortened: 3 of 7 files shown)


romansp:
Fixed no. 2, 3, 4, 5 from #205.

no. 1 - already fixed, just requires localization mentioned in #206.
no. 5 and 6 - can be reproduced only in 'production mode' when `spaceless` filter enabled, so not sure about the fix.

rolandgeider:
Thanks, I'll take a look at this later today

rolandgeider:
About 6 and 7, the only way I see is manually adding a ``&nbsp;`` before the labels, this is kept even after the whitespace filter but it is definitely ugly.

Besides that, I can merge the PR

romansp:
Totally forgot about `nbsp;`. I don't think it is _that way_ ugly :). I will add one more commit with this change, if you don't mind.

rolandgeider:
hehehe, no it's not that ugly, only a bit :)

Sure, go ahead and add the commit

romansp:
Adding `&nbsp;` results in two whitespace characters :cry:. I think the only way to fix it - those _ugly_ `if` clause one-liners with previous HTML tag.

If it's unacceptable I will rebase without this change.

wger/manager/templates/schedule/overview.html
```diff
@@ -20,11 +20,7 @@
     {% for schedule in schedules %}
     <tr>
         <td>
-            <a href="{% url 'manager:schedule:view' schedule.id %}">{{ schedule.name }}</a>
-
-            {% if schedule.is_active %}
-            <span class="label label-primary">{% trans "active" %}</span>
-            {% endif  %}
+            <a href="{% url 'manager:schedule:view' schedule.id %}">{{ schedule.name }}</a>&nbsp;{% if schedule.is_active %}<span class="label label-primary">{% trans "active" %}</span>{% endif  %}
```
rolandgeider: Is it necessary to make the ifs one liners? Shoudn't the nbsp make space in any case? Other than that, the PR is ready

rolandgeider: Ah, sorry, I see it now, without the one liners there are 2 whitespaces.

I would leave the ifs the way they where, the point of the nbsp is just to make a bit of space and this is not so much that it breaks the layout or anything. Plus having a readable template is important as well! :)

romansp:
Rebased and ready.

rolandgeider:
Merged

rolandgeider:
@romansp, since you have an eye for visual stuff, feel free to open issues for bigger problems or things you think should be done differently, many of the pages and overviews have simply grown over time.

PR #223, opened 2015-10-26
Activate localization for share buttons

romansp:
Set `data-lang`attribute, so share buttons will be translated. However, if the current language isn't supported by shariff, it will [fallback to English](https://github.com/heiseonline/shariff/blob/master/src/js/shariff.js#L72).


rolandgeider:
:+1:

Issue #224, opened 2015-10-28
Low Google PageSpeed Insights score


romansp:
Currently http://wger.de gets 51/100 for desktop and 44/100 for mobile https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fwger.de&tab=mobile

![image](https://cloud.githubusercontent.com/assets/3474842/10784615/3d71ce76-7d70-11e5-9013-c8bb3ee08d80.png)

If I submit a PR that improves it up without doing anything weird will you take a look at it?

rolandgeider:
Sure, everything that makes the application faster is appreciated! However there are some things are just not in our hands:

- Browser caching, I recently changed the way wger.de is served and it seems I have forgotten about the caching headers. I'll look into this.
- Reducing the server response time is also difficult, wger.de is in a shared hoster so there is not much I can do there.
- Landing page redirects, this is also something that can't be changed. Once the user arrives at / they are redirected to their language.
-  Adding compression to the site is also that has security problems when [used in combination with https](https://docs.djangoproject.com/en/1.8/ref/middleware/#module-django.middleware.gzip)

These are possible!
- I am however unsure on moving the JS and CSS in above-the-fold. I seem to remember that at least the JS stops working on some pages (probably because what I wrote is... not optimal).
- As for minifying the CSS and JS, the django_compressor app has settings about that, but I haven't looked into it. I think they need some npm modules? On the other hand, the 
- Optimize images

If you need to install a new django app for some of these, go ahead, if it is a well maintained one, there is no problem here.

romansp:
I've submitted PR, that deals with css minification and optimizes several .png files and should improve Google PageSpeed Insights score.

Now, I'd like to try integrate front-end build tools (grunt/gulp), so all front-end optimization stuff will be done on project build automatically. E.g.:
* compile less/scss into CSS (maybe with autoprefixer support)
* compile CoffeeScript, TypeScript, etc. into JavaScript
* concatenate and minify CSS/JS
* minify images
* minify svg
* even pretty evil things like described here https://github.com/addyosmani/critical :smile:
* etc.

And finally generate neat `build` folder where all the generated files will be stored. Afterwards reference those resources in templates.

So, if you don't mind I would like to do it in separate branch and submit another PR.

rolandgeider:
I'd be specially interested in minifying images and svg. I think the concatenation of CSS and JS works alreadz fine with django_compressor (plus it's automatic!) thanks to your other pull request.

[...]

PR #225, opened 2015-10-28
Improve page loading by optimizing front-end resources

romansp:
This PR deals with #224 issue.

Summary:
* optimize several .png files: ~70% less file size, but without actual visual difference.
* enable minification of compressed CSS via `compressor.filters.cssmin.rCSSMinFilter`


rolandgeider:
Looks good. What exactly did you do to optimize the images?

romansp:
Nothing special. I used one of many online services, however most of desktop image editors provide "Save for web" feature. There are also some CLI based tools. Usually they give very close results.

rolandgeider:
OK, I ask because having something more or less automatic would be nice, but since there are not that many pictures that's alright

romansp:
I'll set completely automatic flow for compressing images and svg files via gulp/grunt tasks (I was talking in #224).

Well, if I mentioned it... I would like to open PR implementing this front-end build flow with 'work in progress' label, so you could see my progress. 

Mostly because it will be helpful for me, because I'm not familiar with python/Django stack at all. I think everything will be smooth, but just in case :D

rolandgeider:
Sure, go ahead, nothing will break! :D

PR #231, opened 2015-11-06
Nutriplan

### LINKED ISSUE #210: Nutrition plan: pick correct body weight entry
In the ``get_nutritional_values`` method of ``NutritionPlan`` the per-kg-calculation uses the user's last weight entry. While this is correct when adding meals to the plan, it can be wrong when opening the plan later (the user's weight changed). This should be changed so that the weight entry *closest to the plan's date* should be used.

purplebird:
I don't know what I'm doing.

Possibly fixed #210???

### DIFF
--- a/wger/nutrition/models.py
+++ b/wger/nutrition/models.py
@@ -103,7 +103,7 @@ class NutritionPlan(models.Model):
         '''
         return reverse('nutrition:plan:view', kwargs={'id': self.id})
 
-    def get_nutritional_values(self, closest_weight_entry = None):
+    def get_nutritional_values(self, closest_weight_entry=None):
         '''
         Sums the nutritional info of all items in the plan
         '''
@@ -138,9 +138,9 @@ class NutritionPlan(models.Model):
             for key in result['percent'].keys():
                 result['percent'][key] = \
                     result['total'][key] * ENERGY_FACTOR[key][unit] / energy * 100
-        
+
         # Per body weight
-        if closest_weight_entry != None:
+        if closest_weight_entry is not None:
             for key in result['per_kg'].keys():
                 result['per_kg'][key] = result['total'][key] / Decimal(closest_weight_entry.weight)
 
@@ -150,7 +150,7 @@ class NutritionPlan(models.Model):
                 result[key][i] = Decimal(result[key][i]).quantize(TWOPLACES)
 
         return result
-        
+
     def get_closest_weight_entry(self):
         '''
         Returns the closest weight entry for the nutrition plan.
@@ -159,16 +159,15 @@ class NutritionPlan(models.Model):
         entry = None
         if self.user.userprofile.weight:
             target = self.creation_date
-            closest_entry_gte = WeightEntry.objects.filter(user = self.user) \
-                .filter(date__gte = target).order_by('date').first()
-            closest_entry_lte = WeightEntry.objects.filter(user = self.user) \
-                .filter(date__lte = target).order_by('-date').first()
+            closest_entry_gte = WeightEntry.objects.filter(user=self.user) \
+                .filter(date__gte=target).order_by('date').first()
+            closest_entry_lte = WeightEntry.objects.filter(user=self.user) \
+                .filter(date__lte=target).order_by('-date').first()
             if abs(closest_entry_gte.date - target) < abs(closest_entry_lte.date - target):
                 entry = closest_entry_gte
             else:
                 entry = closest_entry_lte
         return entry
-        
 
     def get_owner_object(self):
         '''
--- a/wger/nutrition/models.py
+++ b/wger/nutrition/models.py
@@ -143,12 +143,14 @@ class NutritionPlan(models.Model):
         if self.user.userprofile.weight:
             # Determine closest entry by date
             target = self.creation_date
-            closest_entry = None
-            for entry in WeightEntry.objects.filter(user = self.user).order_by('date').iterator():
-                if not closest_entry or \
-                (abs(entry.date - self.creation_date) <= abs(closest_entry.date - self.creation_date)):
-                    closest_entry = entry
-            weight = Decimal(closest_entry.weight)
+            closest_entry_gte = WeightEntry.objects.filter(user = self.user) \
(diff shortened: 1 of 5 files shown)


rolandgeider:
Nobody does, @purplebird, nobody does ;)

wger/nutrition/models.py
```diff
@@ -137,10 +138,17 @@ def get_nutritional_values(self):
...
         if self.user.userprofile.weight:
-            weight = Decimal(self.user.userprofile.weight)
+            # Determine closest entry by date
+            target = self.creation_date
+            closest_entry = None
+            for entry in WeightEntry.objects.filter(user = self.user).order_by('date').iterator():
+                if not closest_entry or \
+                (abs(entry.date - self.creation_date) <= abs(closest_entry.date - self.creation_date)):
+                    closest_entry = entry
+            weight = Decimal(closest_entry.weight)
```
rolandgeider: Better than iterating through all the entries, I'd do something [similar to this](https://stackoverflow.com/questions/15855715/filter-on-datetime-closest-to-the-given-datetime). I also think it's possible to replace all the try-excepts by using ``[....].order_by('date').first()``, but I'm not 100% sure

purplebird: Ironically, I actually had it that way originally.

The try-excepts should be unnecessary since we're using less than or equal to and have already confirmed one entry exists with the conditional on 143. Please correct me if you think my logic is off.

rolandgeider: You are right, there will be at least one weight entry when we reach this code.

rolandgeider:
Oh, don't forget to add yourself to the AUTHORS file

purplebird:
Done!

rolandgeider:
I have now taken a look at the errors from travis. Doing a ``.order_by('date')[0]`` doesn't work when the user has no weight entries at all. We'll have to use [first()](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#first), which simply returns None if there are no entries.

As a nice side effect, the tests also check that the correct weight entry is selected in one of the calculations, so we'd just have to update the assertEqual there. The PEP8 formatting errors are easy to fix once you know waht to look for, you can just do a ``pep8 wger`` in the source to check for this.

purplebird:
@rolandgeider , I made a couple sweeping changes here.

I moved the closest entry code into its own function for reusability. I changed the get_nutritional_data() function to now take a weight entry to base the per body weight values. I updated the view to use the new get_closest_weight_entry function for this. I also changed the template so now the user can see which entry the values are based on.

I'm still learning how Django and the template system works, so please forgive small errors. :)

wger/nutrition/templates/plan/view.html
```diff
@@ -186,6 +186,9 @@
 
 
 <h4>{% trans "Nutritional data" %}</h4>
+{% if closest_weight_entry != None %}
```
rolandgeider: You can just do ``{% if closest_weight_entry %}`` like in regular python

wger/nutrition/templates/plan/view.html
```diff
@@ -186,6 +186,9 @@
...
+    <h5>{% trans "Based on the weight entry dated" %} {{closest_weight_entry.date}} ({{closest_weight_entry.weight}})</h5>
```
rolandgeider: Better do here

```
{% blocktrans with date=closest_weight_entry.date  weight=... %}
Based on the weight entry dated {{date}} ({{weight}})
{% endblocktrans %}
```

Then the context is clearer when translating the string

rolandgeider:
Looking good!

BTW, you can run the test locally, just do a ``python manage.py test`` to run the full test suite (you can also do a ``manage.py test wger.nutrition`` to only run the nutrition tests or even ``manage.py test wger.nutrition.test_plan``)

purplebird:
Yeah, I managed to figure that out near the end of my hacking. ;)

wger/nutrition/templates/plan/view.html
purplebird: Thanks.

This looks a lot cleaner.

rolandgeider:
I've thought a bit about this. I think it would be better if ``get_nutritional_values`` would use the closest weight by default, since this is the more "correct" behaviour most of the time. Perhaps something like this? I like the idea of using this to compare differences over time but for the moment the two results we can get ("with weight from back then" and "with current weight") are probably enough. We can expand that in another issue, though. Does this sound OK to you @purplebird ?

```python
def get_nutritional_values(self, use_closest_weight=True):
    if use_closest_weight:
        weight = self.get_closest_weight_entry().weight
    else:
        weight = self.user.userprofile.weight
```

purplebird:
We could do it this way. There is one disadvantage: you would have to ping the database twice to get the weight info for the template (since you're no longer passing it into get_nutritional_values(). Granted, this is relatively minor... Up to you!

rolandgeider:
You are right, but I think we can live with that. On the other hand, we could perhaps return the weight used in the per_kg dictionary. Something like this:
```python
    'per_kg': {'protein': 123,
               'carbohydrates': 123,
               'fat': 123,
               'weight_used': {
                   'weight': 123,
                   'date': 123
               }
    }
```

purplebird:
That is what I originally did, but it explodes where you hit this line:

        # Only 2 decimal places, anything else doesn't make sense
        for key in result.keys():
            for i in result[key]:
                result[key][i] = Decimal(result[key][i]).quantize(TWOPLACES)

And the answer seemed to be either explicitly ignore the 'weight_used' category or do something else here.

rolandgeider:
You are right, lets just fire another query in the view

rolandgeider:
How is it going @purplebird? If you could implement the changes I would merge the PR (I know I have been nitpicking a lot here, but this is the latest change, I promise!)

purplebird:
Sorry!!! I've been distracted working on AI stuff. :)

rolandgeider:
That sounds so cool! Reading and doing some machine learning is also in my personal todo list

purplebird:
Okay. I think that should do it. :)

rolandgeider:
Merged. Thanks again!

PR #233, opened 2015-11-07
Fixed #208

### LINKED ISSUE #208: Warning when opening gym with no members
When opening a gym with no members (e.g. the "default gym" generated by bootstrap_wger), there is an ugly warning by the datatables library because there is no data in the members table

bildschirmfoto am 2015-10-03 um 23 29 10

purplebird:
Kind of a lazy solution for issue #208. Couldn't really figure out why the error was popping up since presumably

    {% for key in user_table.keys %}
        <th>{{ key }}</th>
    {% endfor %}

user_table.keys would be empty without any users.

??


rolandgeider:
This would only remove the table header, which then looks strange. I have now actually read the link in the error and it seems the problem is that datatables can't handle the ``colspan`` here 

```
{% empty %}
<tr>
    <td colspan="4">{% trans "Nothing found" %}</td>
</tr>
```

So it looks like the solution is equally lazy, just remove the block and datables will show its own "table is emtpy" message.

purplebird:
I am not sure what happened here, but I have no problem using "colspan" in the template now, so I think this is fixed. :+1:

rolandgeider:
Mhh, I still get the warning. Better remove it, then we're on the safe side.

BTW, you don't need to merge master into the feature branch, the list of commits gets a bit confusing (there's git rebase in case you want to do it, but I prefer to not use that unless I absolutely have to)

purplebird:
It was 4am when I was doing this and I realized I needed to read more about git but I also needed to sleep...

I thought it would be best to keep the feature branch updated, but maybe this isn't necessary? I wasn't exactly sure what would happen if I tried to open a pull request with code that was behind the master...

rolandgeider:
4am, the best time to code!

Merging in master can be necessary if there have been commits there that affect your changes or touch the same modules, otherwise the worst that can happen are some conflicts during merging and for such small issue fixes this is not a problem. If you would implement some big feature for weeks, then yes, you'd get master into your branch. You can also look into rebasing. Some people insist that all pull requests get *compressed* into 1 commit, but this is something I personally don't like much because I don't like changing the history unless some passwords or similar got commited by mistake.

rolandgeider:
Oh, and yeah, merged

PR #235, opened 2015-11-07
Fixed error when importing duplicate CSV entries

### LINKED ISSUE #204: Duplicate weight entries in CSV import
It seems it's possible to trigger a uniqueness constraint error using the import CSV function for the weight entries. I could have sworn this was already fixed, but it looks it isn't.

During import the view should make sure that duplicate entries are not saved.

purplebird:
This should fix #204.


rolandgeider:
Merged

Issue #241, opened 2015-11-14
Generate nutrition plans with generator


rolandgeider:
The dummy data generator (``extras/dummy_generator/generator.py``) should also be able to generate dummy nutritional plans for users. It should be simple (read: dumb), and just generate 4 meals, for each randomly select an ingredient from the database and a weight between 10 and 250 grams. No need to try to make something realistic here

rolandgeider:
Implemented in PR, closing

PR #245, opened 2015-11-23
Possible fix for issue #241

cthare:
I'm a bit new to working on larger projects and open source in general. I think I may have solved #241.

### DIFF
--- a/extras/dummy_generator/generator.py
+++ b/extras/dummy_generator/generator.py
@@ -448,7 +448,7 @@ if hasattr(args, 'number_nutrition_plans'):
         userlist = [i for i in User.objects.all()]
 
     # Load all ingredients to a list
-    ingredientList = [i for i in Ingredient.objects.all()]
+    ingredientList = [i for i in Ingredient.objects.order_by('?').all()[:100]]
 
     # Total meals per plan
     total_meals = 4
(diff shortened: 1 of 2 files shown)


extras/dummy_generator/generator.py
```diff
@@ -415,3 +437,43 @@
...
+if hasattr(args, 'number_nutrition_plans'):
+    print("** Generating {0} nutrition plan(s) per user".format(args.number_nutrition_plans))
+
+    if args.add_to_user:
+        userlist = [User.objects.get(pk=args.add_to_user)]
+    else:
+        userlist = [i for i in User.objects.all()]
+
+    # Load all ingredients to a list
+    ingredientList = [i for i in Ingredient.objects.all()]
```
rolandgeider: While this is a script that won't get used often, I'm not sure about loading all the ingredients at once. I'd perhaps do something like this (even if this is also not a perfect solution):

```python
ingredientList = [i for i in Ingredient.objects.order_by('?').all()[:100]]
```

extras/dummy_generator/generator.py
```diff
@@ -415,3 +437,43 @@
...
+            nutrition_plan.user = user
+
+            nutrition_plan.save()
+
+            # Add meals to plan
+            order = 1
+            for j in range(0, total_meals):
+               meal = Meal(plan=nutrition_plan, order=order)
+               meal.save()
+               ingredient = ingredientList[random.randint(0, len(ingredientList))]
```
rolandgeider: You can add another for loop here and add a random number of ingredients to the meal, perhaps something between 1 and 5?

rolandgeider:
Thanks!

The PR looks good, I've added some comments, but that's really just nitpicking. Oh, don't forget to add yourself to the AUTHORS file :)

rolandgeider:
I just saw you had added another commit, Github only sends notifications for new messages so I didn't catch it earlier. I've added another comment, but then I can merge this

extras/dummy_generator/generator.py
```diff
@@ -415,3 +437,44 @@
...
+            # Add meals to plan
+            order = 1
+            for j in range(0, total_meals):
+                meal = Meal(plan=nutrition_plan, order=order)
+                meal.save()
+                for k in range(0, random.randint(1,5)):
+                    ingredient = ingredientList[random.randint(0, len(ingredientList)-1)]
+                    i_weight = IngredientWeightUnit(ingredient=ingredient, unit=WeightUnit.objects.all()[0], gram=random.randint(10, 250))
+                    i_weight.save()
+                    meal_item = MealItem(meal=meal, ingredient=ingredient, weight_unit=i_weight, order=order, amount=1)
```
rolandgeider: I see you are generating new weight units, that's not necessary. You can just leave the weight_unit unset (or set it to None) then the system will use grams, for the amount you can just use a random integer. 

```python
meal_item = MealItem(meal=meal, [...] weight_unit=None, amount=random.randint(10, 250))
```

extras/dummy_generator/generator.py
```diff
@@ -415,3 +437,44 @@
...
+
+            nutrition_plan.save()
+
+            # Add meals to plan
+            order = 1
+            for j in range(0, total_meals):
+                meal = Meal(plan=nutrition_plan, order=order)
+                meal.save()
+                for k in range(0, random.randint(1,5)):
+                    ingredient = ingredientList[random.randint(0, len(ingredientList)-1)]
```
rolandgeider: Oh, and here you can also just do

```python 
ingredient = random.choice(ingredientList)
```
that's a bit cleaner

cthare:
I actually meant to send a message with the previous commit, but I had something come up right after I submitted. I just made the other changes to remove the creation of a weight unit and cleaned up the ingredient line.

rolandgeider:
Merged, thanks again.

PR #250, opened 2016-01-08
Multiple choices checkboxes. This fixes #244.

### LINKED ISSUE #244: Render multiselects as checkboxes
How to use the multiselect boxes is not always clear, who knows that it's necessary to do Click + Ctrl? (mobile is in this aspect better because mobile browsers render it differently)
checkbox

It should be possible to render this as a list of checkboxes that can be selected individually. While django has a ``CheckboxSelectMultiple`` that does this, the HTML is incompatible with the bootstrap classes so the result looks really ugly. The problem in this issue is finding a solution for this. There are some django apps to render forms with the bootstrap classes, if they are needed for this to work, that's OK too.

murphyalexandre:
Created a custom widget and renderer to make it reusable.

### DIFF
--- a/wger/core/templates/tags/render_form_element.html
+++ b/wger/core/templates/tags/render_form_element.html
@@ -12,9 +12,9 @@
         </label>
 
         <div class="col-md-9">
-            {% if not field|is_checkbox and not field|is_fileupload %}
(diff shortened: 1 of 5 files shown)


rolandgeider:
Awesome! This had been itching me a long time

Issue #64, opened 2014-02-08
More flexible logs


rolandgeider:
When adding a log, either in the log view or in the gym view, it should be possible to exchange one exercise with another. This can be useful e.g. when a machine is occupied or the user can't do the planed exercise for other reasons.

rolandgeider:
From user feedback: it should be possible to skip exercises. This makes sense, it should be possible to quickly navigate to a specific exercise (or skip it, etc.)

rolandgeider:
Add explicit "on the fly workout" option, so that there are no initial exercises and the user can select on each step what they want to do

rolandgeider:
Other ideas:

* Show the exercise images, if any.
* Show the exercise description and tips
* When selecting the repetitions, use a dropdown, not an input field, this is faster in mobile devices
* Even better, since the user has probably already done that specific exercise weight-repetions combination, the page could present somehow the last ones, when the user selects one, both the repetitions and the weight are filled in.

rolandgeider:
Adding the needs dicussion tag, since it's not clear (to me) exactly how some of these ideas could actually look like (if they are useful at all)

rolandgeider:
From a user request: should the timer page reload for whatever reason (this can happen more easily on the mobile version), it should be possible to resume the session.

dometto:
> Add explicit "on the fly workout" option, so that there are no initial exercises and the user can select on each step what they want to do

:+1:

Other apps (like thesquatrack.com) also allow adding exercises to a day that are not coupled to a routine, "workout day", or whatever. This is essential to some users. I haven't looked at the code yet, but I would be interested in contributing to this if it wouldn't require a complete rewrite of the app. :)

rolandgeider:
hi @dometto !

I know, there have been many days where I wish I could do this! The positive side is that it's not necessary to rewrite the whole application, the backend can already support this, what's needed is to make the frontend more flexible. If you want the technical details: currently the gym mode is made with jquery, we plan to add angularJS to the application to allow for more dynamic pages (starting with #284, but this issue would be the second thing to attack). It wouldn't make much sense to add now the flexible logs and then rewrite everything and re-implement it again. If you are fit with angular you'd be more than welcome to contribute, @illegalnumbers is on board as well

dometto:
It's been a while but am interested in using wger again -- what is the general status of the project, and of this issue (and the angular implementation) in specific?

dometto:
Was this feature introduced, and can I test it anywhere?

[...]

PR #252, opened 2016-01-12
Angular timer to fix #171

### LINKED ISSUE #171: Use angularJS for timer page
The timer page is the perfect example of a SPA, instead of the current JS mess, this should be rewritten with angularJS. There is a current work in progress in the angular_timer branch.

Implementing this issue would also pave the road to fixing #64

murphyalexandre:
I'm opening this up so we can discuss the issue and the code that will fix it.

I'm guessing you want a countdown running up for the exercises and down for the pauses in the bottom of the timer screen, am I right ?


rolandgeider:
Thanks for opening this.

As a first step I just wanted to "port" the current logic to angular without any new logic. Once that is done, we'd take a look at #64 and see which ideas make sense and implement those. This is mainly just to prevent feature creep, something I'm often guilty of doing, but wanted to avoid here.

murphyalexandre:
@rolandgeider Mind taking a look and telling me if I'm missing anything so I can add everything before styling it up ?

rolandgeider:
Looking good! I like the slight animation on the progress bar. The only thing missing is that after finishing the last step, you should have a dialog to save your impressions, the API endpoint is ``workoutsession`` (the start time and end time are set internally, not by the user)

bildschirmfoto am 2016-01-14 um 10 43 47

murphyalexandre:
I added the last page, but it still needs some love, like keeping a constant timer and using a dropdown for the impressions.

Issue #272, opened 2016-03-05
Extended PDF options for schedules


rolandgeider:
The PDF generator for schedules should be extended so that it has the same options as the regular workout (images, comments, etc.). This should be fairly easy, since most of the logic and code from the workout view, urls, etc. will work as well

alelevinas:
i'm on it :)

rolandgeider:
Great! :)

rolandgeider:
PR merged, closing here

PR #290, opened 2016-05-20
Extended Schedules pdf generator  for issue #272

alelevinas:
This extends the pdf options for schedules so it is equal to the workout options for pdf downloading.

Issue #272 

I continued the pdf export in the schedule.py file. Do you want everything in the pdf.py file?

Let me know if everything is ok.

### DIFF
--- a/wger/manager/templates/schedule/view.html
+++ b/wger/manager/templates/schedule/view.html
@@ -251,17 +251,7 @@
         {% trans "Download as PDF" %}
     </a>
 </p>
-<!--
-<p>
-    <a href="{% url 'manager:schedule:pdf' schedule.id uid token %}">
-        <img src="{% static 'images/icons/save.svg' %}"
-             width="32"
-             height="32"
-             alt="{% trans 'Download as PDF' %}"
-             title="{% trans 'Download as PDF' %}">
-        {% trans "Download as PDF" %}</a>
-</p>
--->
+
 {% if is_owner %}
 <p>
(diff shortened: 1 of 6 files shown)


rolandgeider:
Thanks, I'll take a look this weekend

alelevinas:
sorry for that. I only run the tests for the manager.schedule part.
I'm resolving the other tests right now

rolandgeider:
No worries, it happens to me regularly as well. I didn't get to take a closer look, had problems with the internet connection and just got it working again.

wger/manager/templates/schedule/view.html
```diff
@@ -226,51 +226,53 @@ <h3 class="panel-title">
 <h4>{% trans "Adding workouts" %}</h4>
 
 <p>{% blocktrans %}Add as many workouts to the schedule as you want. You can
-edit the order by dragging and dropping them. It's also possible to add one workout more
-than once.{% endblocktrans %}</p>
+    edit the order by dragging and dropping them. It's also possible to add one workout more
```
rolandgeider: Even if it doesn't look nice, better don't format the texts within blocktrans, when I extract the strings for translations, they are then marked as new (because there are more spaces). Then, even if then transifex offers a suggestion to use the old translation, this is a bit more work for the translators.

(Just for the future, I'll just change these back)

rolandgeider:
Looks good, merged :+1:

PR #300, opened 2016-07-01
add clarification on where to install Bower

jstoebel:
Bower should be installed in wger/wger not in the project's root directory


rolandgeider:
Thanks! Just merged this

PR #312, opened 2016-08-16
Deprecation warning.

petervanderdoes:
When starting the server a deprecation warning is given:

```
wger/source/wger/urls.py:149: RemovedInDjango110Warning: Support for
string view arguments to url() is deprecated and will be removed in
Django 1.10 (got django.contrib.sitemaps.views.sitemap). Pass the
callable instead.
  name='sitemap')
```


rolandgeider:
Thanks... and merged :)

PR #313, opened 2016-08-16
Deprecation warning context_processors

petervanderdoes:
A warning message is shown when starting the server:
```
wger/source/wger/utils/generic_views.py:24: RemovedInDjango110Warning:
django.core.context_processors is deprecated in favor of
django.template.context_processors.
  from django.core.context_processors import csrf
```


rolandgeider:
Merged

PR #315, opened 2016-08-19
Fix django.core.context_processors deprecation.

petervanderdoes:
Missed several deprecated imports during the previous fix.


rolandgeider:
Thanks, merged

Issue #309, opened 2016-08-06
BMI calculator graph does not change with weight unit setting


SPCOxion:
The graph on the BMI calculator page is always in kilograms, even if you change your weight unit in preferences.

For example, I am 5'7" (approx. 170cm) and I weigh 140lbs. When I input my height and weight, it gives me an accurate BMI (21.97) inside the healthy range. However, the graph shows my height and height in the "Adipositas III" range. (image 1)

Similarly, when I input my weight in killograms (approx. 63kg), it gives me an inaccurate BMI (9.89), but the graph shows me in the "Normal Weight" range. (image 2)

Image 1:
![2016-08-06 00_37_12-bmi calculator](https://cloud.githubusercontent.com/assets/3937393/17454941/12cc95ac-5b6e-11e6-83d7-6da4a877bd23.png)

Image 2:
![2016-08-06 00_38_47-bmi calculator](https://cloud.githubusercontent.com/assets/3937393/17454942/3ec1756a-5b6e-11e6-80e0-318c552e92f7.png)

petervanderdoes:
The Y-Axis in the graph is set to represent KG. The 140lbs is sent to the graph is 140 kg, hence the wrong dot location.

petervanderdoes:
After PR #314 is approved and merged, I can make the chart show lbs when the personal preference is lbs.

rolandgeider:
PR was merged which fixed the issue, closing it.

@SPCOxion : I won't be able to update wger.de till next week

PR #318, opened 2016-08-19
Bugfix/bmi chart in kg only

petervanderdoes:
When the user has their setting set to lbs, the BMI chart is displayed incorrectly. The chart shows Kg only.

This PR gets the data for the x and y axis from the data send through the JSON request.
The send JSON data is send in Kg or Lbs depending on the users settings.


petervanderdoes:
This will solve issue #309

rolandgeider:
Thanks, you're on fire! :+1:

PR #325, opened 2016-09-04
Fix height calculation

petervanderdoes:
Height calculation for the BMI graph is incorrect.


rolandgeider:
Thanks for fixing this! When merging such issues, don't forget to add a reference in the changelog, e.g. like 07b8109, for small things there's no need to mention anything, just list them

petervanderdoes:
Ok :+1:

PR #360, opened 2017-02-23
Bugfix/missing gulp

petervanderdoes:
With Python 3.6 in Travis gulp was missing. In hindsight I don't understand why it worked for other versions. It's all fixed now.

This Pr also has Python 3.6 testing in Travis.
Updated NPM and updated Node.


rolandgeider:
Thanks!

petervanderdoes:
The develop branch is a local branch based of the master branch. I use the git-flow workflow, meaning master is for production and development, well is for development. :) This merge is to update after pr #372 Use newer setuptools

rolandgeider:
I have always thought that git-workflow is a bit too big and only makes sense if there are old releases that are developed or maintained in parallel to current ones. Do you use this on smaller things?

petervanderdoes:
I use it for all projects I work on. Two main reasons I use it; uhmm i created git-flow avh, a more up to date version of git-flow and second it keeps the production, develop, and release candidates separated.

Say I'm working on a program and the release is 1.0.0. Right now I would be working on the next release, implementing new features etc. Then somebody discovers a security flaw in 1.0.0. I can't just fix it in the branch I'm developing on as it would introduces all whole bunch of new features that are not tested it. So I create a branch of the production branch (gitflow term: hotfix), fix the flaw, merge it back into the production and the develop branch, release a new version based of the new production branch.

Final thing I want to say, git-flow isn't so much about multiple old releases being maintained, it's more about separating production and development.

rolandgeider:
thanks for the explanations, I'll give git-flow avh a try

petervanderdoes:
I think the milestone for this needs to be pulled up to 1.8 as every push to the repository fails on the Travis-CI check because of the bug this PR fixes.

rolandgeider:
1.8 was released yesterday, this can be merged now

PR #369, opened 2017-03-27
Add birthdate field to userprofile

dtopal:
-user can change from preferences
-gym admins can enter birthdate when creating user


rolandgeider:
Awesome! I'll take a closer look as soon as I can

dtopal:
Thanks!  Sorry, I'm still relatively new at this.  I'm a bit lost as to why it's failing the check now but passing the check locally.

rolandgeider:
It seems we need a newer version of setuptools, there's already PR for that #372

dtopal:
Ok, I'll pull in the changes and see if that works.

dtopal:
Should I clean up the commit history?

rolandgeider:
No need to, I am also not a big fan of constantly rebasing. I mean, if you want to feel free, but unless there are _a lot_ of unecessary commits, you can just leave them

rolandgeider:
I'm currently finishing the 1.8 release, when that's over I'll take a look at his (but the first impression is good!)

dtopal:
Thanks!

wger/core/signals.py
```diff
@@ -40,5 +43,17 @@ def create_user_cache(sender, instance, created, **kwargs):
...
+def set_user_age(sender, instance, **kwargs):
+    '''
+    Inputting or changing a user's birthdate will auto-update the
+    user's age as well if user's age has not been set
+    '''
+    if instance.age is None and instance.birthdate is not None:
+        today = datetime.date.today()
+        birthday = instance.birthdate
+        instance.age = (today.year - birthday.year
+                        - ((today.month, today.day) < (birthday.month, birthday.year)))
```
rolandgeider: Shouldn't this be ``(month, day) < (bday.month, bday. day)``?

wger/core/models.py
```diff
@@ -207,6 +222,14 @@ class UserProfile(models.Model):
                        validators=[MinValueValidator(10), MaxValueValidator(100)])
     '''The user's age'''
 
+    birthdate = models.DateField(verbose_name=('Date of Birth'),
+                                 help_text=_('Please use the format: mm/dd/yyyy'),
```
rolandgeider: Django should be clever enough to adapt the format to the current locale so if browse the application in German you'd enter ``dd.mm.yyyy``. Also if datefields are rendered as date input fields (I'm not 100% sure that they are), the browser will validate the format as well

wger/gym/forms.py
```diff
@@ -65,6 +65,9 @@ class GymUserAddForm(GymUserPermisssionForm, UserPersonalInformationForm):
     Form used when adding a user to a gym
     '''
 
+    birthdate = forms.DateField(required=False,
+                                help_text=_("yyyy/mm/dd"))
```
rolandgeider: Same as above, we can just remove the help text

wger/core/signals.py
dtopal: Yes it should, thanks!

wger/core/models.py
dtopal: Ah, I see.  I'll get to work on the changes.

dtopal: Sorry, I have a quick question.

Should I be using the Html5DateField from wger/utils/fields.py for the bday field in the model and the Html5FormDateField from wger/utils/widgets.py for the form field?

rolandgeider: Better not, it seems there are still some browser compatibility issues with date input elements, just leave it as it is

dtopal:
I think that's all the changes.  Let me know if I missed something or if there are any other changes you think would be good.

PR #380, opened 2017-04-16
avoid weird curve in weight graph

alokhan:
The weight graph was sometimes really weird this solve the problem.

image


rolandgeider:
Thanks for the PR, is there a reason you closed this? :)

alokhan:
I had fetched all the changes from upstream and i didn't want it to be in the PR :) I created a new one.

PR #381, opened 2017-04-16
solve weird weight graph

alokhan:
Solve the weight graph being weird sometimes see screenshot

image


rolandgeider:
Thanks, just merged this. Will probably update wger.de this weekend

PR #384, opened 2017-04-22
Easier to understand installation process description.

mpena2099:
Easier to understand installation process description (for ME, at least). From https://wger.readthedocs.io/en/latest/development.html documentation.


petervanderdoes:
As you noticed yourself, see issue #385, these instructions actually don't work.

PR #390, opened 2017-05-01
Bugfix/gym user permisssion form

petervanderdoes:
Three changes:

- Replace mutable default argument .
- Editing of roles doesn't work
- Rename the class, just a small typo.


rolandgeider:
Changes already merged with speshaks PRs

PR #391, opened 2017-05-01
Errors running the generator.

petervanderdoes:
When running the generator it can't find the settings file.


rolandgeider:
The generator is now split into individual management commands, this is not needed anymore

PR #398, opened 2017-05-23
Suggested fix to Dockerfile

Tajmahal86:
Tried to build the Dockerfile and ran into some Not Found erros while trying to apt-get install.

Updating the repositories with apt-get update solves this.

### DIFF
--- a/extras/docker/apache/Dockerfile
+++ b/extras/docker/apache/Dockerfile
@@ -21,7 +21,7 @@ ENV LC_ALL en_US.UTF-8
 
 
 # Install dependencies
-RUN apt-get install -y apache2 libapache2-mod-wsgi-py3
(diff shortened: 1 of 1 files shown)


DandyDeveloper:
The only reason I'd be hesitant of including this as a fix is because future dependencies could change and introduce new problems. 

It might be better to just update the base image for now with the current updates to prevent this. 

@rolandgeider's Call though!

rolandgeider:
We could change the base image ubuntu version to 17.04, this would also 
ensure that the wger images don't get too big

Am 07.10.2017 um 19:57 schrieb Aaron Layfield:

rolandgeider:
The dockerfiles have been changed a lot in the meantime. Thanks anyway for this PR!

PR #405, opened 2017-08-02
Fix docker/apache docker API access

withnale:
At present the API doesn't work when using the docker wger/apache image.

This is because the ``WSGIPassAuthorization On`` apache directive isn't specified.

This can be found on the rest_framework documentation at the following URL

http://www.django-rest-framework.org/api-guide/authentication/#apache-mod_wsgi-specific-configuration

### DIFF
--- a/extras/docker/apache/wger.conf
+++ b/extras/docker/apache/wger.conf
@@ -9,6 +9,7 @@
     WSGIDaemonProcess wger python-path=/home/wger/src python-home=/home/wger/venv
     WSGIProcessGroup wger
     WSGIScriptAlias / /home/wger/src/wger/wsgi.py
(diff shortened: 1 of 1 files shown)


DandyDeveloper:
Sorry for the late response on this. Just checking, have you tested this to make sure its the only WSGI setting needed for Apache deployments? 

Change is definitely needed, just wanted to make sure nothing else is missing in the Apache config.

volkanx:
@withnale Thanks for this. I'm having the same issue. I've updated wger.conf and restart the container but still getting the same issue. Do you know if there's anything else that needs to be done for the already installed applications?

Thanks.

rolandgeider:
This was fixed in #410 in the meantime, closing here

PR #412, opened 2018-01-07
Adds configuring of allowing users to add new exercises and ingredients - fixes #382

### LINKED ISSUE #382: Allow configuring adding new ingredients and exercises
From user feedback: it should be possible to configure the application so that users can add exercises or ingredients directly, without going through an administrator, this can be a new variable in the config app. We can have different settings (for each the ingredients and the exercises), e.g.

- users can't submit anything
- users submit things and an admin has to check (current behaviour)
- users submit directly, admins just get notified

dtopal:
Adds settings to let users add ingredients and exercises without going through an administrator.  Also allows user settings to be changed so that:
- users can't add anything
- users can submit new exercises and ingredients for approval (current behavior)
- users can submit new exercises and ingredients without needing approval

fixes #382

### DIFF
--- a/wger/config/forms.py
+++ b/wger/config/forms.py
@@ -28,4 +28,5 @@ class UserCreateItemPermForm(forms.ModelForm):
     class Meta:
         model = UserCanCreate
         fields = ('ingredient_perm',
+                  'exercise_perm',
                   )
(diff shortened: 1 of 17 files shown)


dtopal:
gulp isn't launching in the Travis CI build.

@rolandgeider is it possible that the wger/.travis.yml file needs to updated?
Currently it says: npm install gulp gulp-cli

Should it be "npm install -g gulp-cli" instead?  I'm not sure.

PR #421, opened 2018-03-13
Added external libraries version

Veltarn:
I added versions on some externals library because some of them changed their directory structure resulting on HTTP 404 when loading these libraries.


rolandgeider:
Thanks!

PR #437, opened 2019-01-06
fix(static): get profile data from directly from userprofile object

Aerex:
*Context*: When trying to show or hide comments a syntax error show up in the development console 

Application does not crash but the toggle option to show or hide user comment does not persist 

wger_core


rolandgeider:
This was fixed in the meantime. Thanks anyway!

PR #438, opened 2019-02-01
apache fqn fix

felipegouveiae:
I had some issues to run the docker image wger/apache

I added an apt-update to fix apache2 package dependences and added a commando do register localhost as fqn on apache. 

There's an error building the image : 

django-recaptcha 2.0.2 has requirement django<2.3,>1.11, but you'll have django 1.10.8 which is incompatible.

We have a problem because in order to build the image we need to get the requirements txt fixed to the image download the code and build the image properly.

### DIFF
--- a/extras/docker/apache/Dockerfile
+++ b/extras/docker/apache/Dockerfile
@@ -21,7 +21,10 @@ ENV LC_ALL en_US.UTF-8
 
 
 # Install dependencies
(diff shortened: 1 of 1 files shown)


rolandgeider:
Thanks for the PR! Tthis change was included in #458 (possibly pulled from here?), closing here

PR #442, opened 2019-05-08
Added directive to pass api auth token to wsgi

sahoahfoa:
Apache will not pass the authorization token to the application which is needed to access private api endpoint.


rolandgeider:
Thanks for the PR, this change was included in #457 (possibly pulled from here?), closing here

PR #443, opened 2019-05-09
Peg dependencies to django 1.10 compatible versions

speshak:
django-filter & django-cors-headers didn't have an explicit version and
newer versions were causing docker image build failures.  These versions
are based on what is currently in the published wger/apache docker
image.


speshak:
I ended up also changing how gulp is installed in TravisCI in order to get the builds to pass.

PR #457, opened 2020-06-10
A collection of patches, mainly related to Docker improvments

speshak:
This is the first in a series of PRs that I intend to open to get work that I've done modernizing wger into the proper project.  This PR is primarily changes to the Docker build process.  It also includes patches from @Aerex and @Withnale which were lifted from their respective forks.


speshak:
Part of #459

Issue #462, opened 2020-06-14
possible feature ? BMI on Graph 


sinedo:
Although the BMI is not as good as other measurements it gives you a rough estimate where your weight should be located 
It would be cool to have a checkbox which enables the a transparent color of the bmi displayed in the background of the graph 

![Graph_with_BMI](https://user-images.githubusercontent.com/46466615/84596315-3a358900-ae5d-11ea-9c04-2be4cfb7e16f.png)

rolandgeider:
Yeah, this sounds useful

Svn-Sp:
@rolandgeider I'm new to Github, but I'd like to try this one.

rolandgeider:
Hi @Svn-Sp ! Sure, do you have a working development instance? I still haven't uploaded the new Docker images... 🙈 We can coordinate us in the gitter channel, or here if you want

Svn-Sp:
Have installed wger for developement, but when I go to the weight overview page I just get the table but no graph...

rolandgeider:
Strange, if you have entries there, it should show you something. Does the development server show 404 errors?

Svn-Sp:
![screen](https://user-images.githubusercontent.com/69042800/89727340-04a7b900-da24-11ea-850b-d158c40486a0.jpg)
There are no 404's in the console...

rolandgeider:
I'll take a look at master. I've been working on a different branch with mobile templates (which I'll hope I'll be able to merge soon) and there it's definitely working

rolandgeider:
I've merged the branch into master and the weight overview is working (other things are ugly, but should still be functional). The CSS/JS libraries are downloaded with yarn now, the easiest thing is probably to use docker-compose, but you can download the dependencies yourself with ``yarn install``

Svn-Sp:
Excuse me if this isn't the smartest question, but could you post the complete ```yarn install``` command?

rolandgeider:
``yarn install`` is the whole command 😄. You call it from the source folder and it should download everything to wger/core/static/yarn/

![image](https://user-images.githubusercontent.com/2280729/90024783-924e0780-dcb5-11ea-988d-a462b064654f.png)


I really can't think of a reason why it wouldn't show anything

Svn-Sp:
Still the same problem; No graph is showing and no 404's in the console.
I think ```yarn install``` did not work properly
[...]

PR #497, opened 2020-08-19
Feature/bmi on graph at weight overview (Issue #462)

Svn-Sp:
Maybe the most hacky thing I ever did... 
There is no inbuilt function for custom ticks, so I had to make the code calculate the correct positions of the bmi lines manually.
The only thing I have not figured out yet is how to give the lines a different color. I do not know why, but ```setAttributeNS(null, "stroke", "red")``` does not work...


rolandgeider:
> Maybe the most hacky thing I ever did...

OMG you weren't kidding! 😂

I'll take some time tomorrow or so to look at it

rolandgeider:
I've been looking around the docs of metricsgraphics (which doesn't seem to be maintained so well at the moment BTW) for a graph for the new nutrition diary. Wouldn't it be enough to set a couple of baselines? While we can't set the color, this would be cleaner and easier for us. What do you think?

image

Svn-Sp:
Could you send a link please?

rolandgeider:
Sure!

https://github.com/wger-project/wger/blob/46f416ad0520a3db355f928b972cb2e84147bf56/wger/nutrition/static/js/nutrition.js#L114

https://github.com/metricsgraphics/metrics-graphics/wiki/Graphic#wiki-baselines

rolandgeider:
Did you manage to take a look?

Svn-Sp:
Had a really busy week, I will take a look at it now!

Svn-Sp:
Finished it! Much faster than last time XD.
Could not solve the merge conflict at AUTHOR.rst, but I am not that much of an Github expert, so it should be no problem to solve it.

rolandgeider:
PR merged, sorry it took so long 😅

PR #510, opened 2020-09-04
Nutrition Journal chart type change

### LINKED ISSUE #501: Use bars instead of lines for nutrition diary chart
Currently the chart for the nutrition diary is a line chart. Since users probably won't be logging everything every single day, a bar chart would be better. While metricsgraphics does support this, changing the type of chart causes some NaN error somewhere in the JS stack.

Additionally, it seems that metricsgraphics is not developed much at the moment and might have to be replaced. While replacing it would go beyond the scope of this issue, it's OK to discuss it if no other solution is possible

image

WalkingPizza:
## Proposed Changes

  - Changed the Nutrition Journal chart to a bar chart
  - Fix #501

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -26,6 +26,7 @@ Developers
 * Boniface Mwenda: https://github.com/andela-bmwenda
 * Scott Peshak: https://github.com/speshak
 * Musanje Louis Michael: https://github.com/louiCoder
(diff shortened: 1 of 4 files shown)


rolandgeider:
I could have sworn I tried that! 😅

In any case, thanks!!!

PR #512, opened 2020-09-04
Support latest dev release (local/remote) with entrypoint

ivandir:
### Proposed Changes

  - Allows for docker-compose deployment of latest development dockerfile.
  - Docker volumes now persist as named volumes
  - Entrypoint script is being called directly since 2.0 tag doesn't work

- [ ] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8``)
and isort (``isort``) 

### Other questions

this PR?) No just allows for docker-compose deployment of latest development dockerfile.

### DIFF
--- a/extras/docker/compose/docker-compose.yml
+++ b/extras/docker/compose/docker-compose.yml
@@ -2,11 +2,10 @@ version: '3.7'
 
 services:
   web:
-    image: wger/devel:2.0-dev
+    image: wger/devel:latest
+    command: bash -c "/home/wger/entrypoint.sh"
     volumes:
-      - type: bind
-        source: ../../../
-        target: /home/wger/src/
+      - 'wger:/home/wger/src'
     ports:
       - 8000:8000
     env_file:
@@ -34,3 +33,4 @@ services:
 
 volumes:
   wger-postgres-data:
+  wger:


ivandir:
@rolandgeider this is my first PR, please let me know if I missed something.

kuggaa:
I need to pay someone for install for special needs kids exercise - tea@bellsouthwest.com 



Non secure message

rolandgeider:
What do you mean that the 2.0 tag doesn't work? The code is automatically built and tagged with 2.0 and latest

rolandgeider:
How do you develop with this setup? Do you open the volume location directly?

rolandgeider:
I just thought, if we want to persist stuff, we should also put the media folder in a volume

ivandir:
> What do you mean that the 2.0 tag doesn't work? The code is automatically built and tagged with 2.0 and latest

Not sure what was wrong with the 2.0 tag but I'll try it again. It wouldn't run the bootstrap steps correctly and for some reason it skipped migrations and fixtures.

ivandir:
> How do you develop with this setup? Do you open the volume location directly?
I rebuild the image and relaunch the container.

rolandgeider:
> > How do you develop with this setup? Do you open the volume location directly?
> I rebuild the image and relaunch the container. I see what you're saying though, it makes more sense if you're developing within to just keep the building as so:
> 
> ```
> volumes:
>   - type: bind
>     source: ../../../
>     target: /home/wger/src/
> ```

I think it will be better to make a new docker compose that is only updated with new images and leave this as it is for development

rolandgeider:
I've started a new repo with only the docker-compose files to use for production-like environments: https://github.com/wger-project/docker People that don't want to develop, can just pull that and get started. Closing this PR here

Issue #508, opened 2020-09-01
Use f strings for formatting


rolandgeider:
Now that we use python 3.6+, we can probably replace and simplify many of the string formatting and replace them with [f-strings](https://realpython.com/python-f-strings/) where useful

(yes, this is purely for aesthetic reasons)

PR #530, opened 2020-10-01
Convert /core to fstrings

sztang:
Addressing #508 ; converted all instances of string format method to fstrings in the /core app. Can't change the strings that are inside _() ugettext functions though, since it would interfere with the translation.


rolandgeider:
Awesome!!

sztang:
Thanks for merging! Will do more fstring cleanup on the other apps once I can find time.

PR #566, opened 2020-12-02
Exercise variations

ryowright:
### Proposed Changes

  - Added many-to-many field to exercise model for exercise variations
  - Added display of exercise variations in frontend exercise detail view
  - Created exercises_with_variations.json fixture for initial data load

and isort (``isort``) 

### Other questions

this PR?)

Adds new "variations" field to Exercise model in databases. Users may need to load the data from the exercises_with_variations.json fixture if it is not loaded already.


wger/exercises/migrations/0013_auto_20201202_0043.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.DeleteModel(
+            name='ExerciseVariation',
+        ),
+        migrations.AlterField(
+            model_name='exercise',
+            name='variations',
+            field=models.ManyToManyField(blank=True, related_name='_exercise_variations_+', to='exercises.Exercise', verbose_name='Variations'),
+        ),
+    ]
```
rolandgeider, line 21: That's a lot of migrations 😅

You can go back to the last one with `python3 manage.py migrate exercises 0005`, delete the new ones and then generate a single one with all the changes

wger/exercises/templates/exercise/view.html
```diff
@@ -253,6 +253,18 @@ <h5 class="mt-4">{% trans "Weight log" %}:</h5>
...
+    {% for variation in exercise.variations.all %}
+        <li>
+            <a href="{{ variation.get_absolute_url }}">{{ variation }}</a>
+        </li>
+    {% endfor %}
+    </ul>
+{% else %}
+    <p>{% trans "None" %}</p>
+{% endif %}
 {% endblock %}
```
rolandgeider, line 268: I'd wrap the whole section in the {% if %} and not show anything if there aren't any variations

rolandgeider (changes requested):

wger/exercises/models.py
```diff
@@ -153,6 +153,24 @@ def delete(self, *args, **kwargs):
...
+class Variation(models.Model):
+    """
+    Variation ids for exercises
+    """
+
+    def __str__(self):
+        """
+        Return a more human-readable representation
+        """
+        return self.name
```
rolandgeider, line 165: The variation doesn't have a name, better put something like return f'variation {self.id}' here

wger/exercises/views/exercises.py
```diff
@@ -133,8 +133,11 @@ def view(request, id, slug=None):
     template_data['show_shariff'] = True
 
     exercise = get_object_or_404(Exercise, pk=id)
+    variations = Exercise.objects.select_related().filter(
+        variations=exercise.variations).exclude(name=exercise.name)
```
rolandgeider, line 137: Since we will probably need the variations in other places, I would suggest moving this filter logic to the exercises model, perhaps a property like this?

```python
@property
def get_variations(self):
    return self.variations.exercise_set.exclude(id=self.id)
```

then we can just access it in the template

rolandgeider (changes requested):

rolandgeider (approved):

Issue #426, opened 2018-04-27
Add minimum length to ingredient name


Wieke:
So there appears to be a minimum term length (3 characters) for the ingredient search but no minimum length for ingredient names. So it's possible to create an ingredient that can't be found because the name is too short. Granted there are probably not that many ingredients with such a short name, though the default list of Dutch ingredients contains "Ui" (onion). 

The obvious solution to me seems to add minimal ingredient name length to the ingredient creation forms.

austin-leung:
I'll look into this!

rolandgeider:
👍🏻👍🏻

PR #569, opened 2020-12-05
Add minimum length to Ingredient name (Issue #426)

austin-leung:
### Proposed Changes

  - Change minimum required length for an ingredient name to be 3 (affects create/edit)

- [X] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8``)
and isort (``isort``) 

### Other questions

this PR?)
No

This is what it looks like now if you try to create/edit an ingredient name of too few characters:
image

One thing I'm not sure about is whether I should write tests for this. I see tests around the Ingredient model in `wger/nutrition/tests/test_ingredient.py`, but I don't think we currently test other validators such as `MinValueValidator`

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -32,6 +32,7 @@ Developers
 * Biplov - https://github.com/beingbiplov
 * Rashmi Kamath - https://github.com/Rkamath2
 * Anthony (Ryo) Wright - https://github.com/ryowright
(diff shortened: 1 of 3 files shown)


rolandgeider:
You are right, I don't think we test many other validators. I mean, if you want to add some tests that wouldn't be wrong of course, but it's also not necessary

rolandgeider:
Awesome! 😎

Issue #568, opened 2020-12-03
Add kilojoule values to ingredient


rolandgeider:
While the ingredients store the energy in kcal, it would be nice to also show the value in kilojoule. This can be accomplished easily with a property in the ingredient model.

1 kcal are 4,184 kJ

rolandgeider:
Hi! Sure, go ahead

nopinter:
Hi, my partner, @derekli17, and I are also looking into this!

JuanFML:
Hi! Sure go ahead I am a little bit stuck with creating the property in _wger\nutrition\views\ingredient.py_, although I got to edit the view in _\wger\nutrition\templates\ingredient\view.html_

PR #577, opened 2020-12-13
Added kilojoules property to ingredient model (Issue #568)

nopinter:
### Proposed Changes

  - added "energy_kilojoule" property to Ingredient class in wger/nutrition/models.py

  - added kilojoules to display on ingredient details page in wger/nutrition/templates/ingredient/view.html

  - added to wger/nutrition/static/js/nutrition.js, making kilojoules responsive to updates in ingredient amount according to user input

  - added test to wger/nutrition/tests/test_ingredient.py, verifying correct kilojoule conversion

- [✓] Tests for the changes have been added (for bug fixes / features)
- [✓] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8``)
and isort (``isort``) 
- [✓] Added yourself to AUTHORS.rst

### Other questions

this PR?) 

No


nopinter:
I worked on these changes with my partner, @derekli17, which is why his name is also included in AUTHORS.rst. We are currently undergraduates studying computer science, so if you have any feedback or want us to make changes, please feel free to let us know! Thanks.

wger/nutrition/models.py
```diff
@@ -512,6 +512,17 @@ def get_owner_object(self):
         """
         return False
 
+    def _get_energy_kilojoule(self):
+        """
+        returns kilojoules for current ingredient, 0 if energy is uninitialized
+        """
+        if self.energy:
+            return self.energy * 4184
```
rolandgeider, line 520: Sorry, 1kcal ist 4.1 kJ, not four thousand! :D I had just copy pasted it from the German wikipedia and hadn't thought that not everybody uses commas for this

wger/nutrition/models.py
```diff
@@ -512,6 +512,17 @@ def get_owner_object(self):
...
+        else:
+            return 0
+
+    energy_kilojoule = property(_get_energy_kilojoule)
```
rolandgeider, line 524: You can just define the property directly

```python
@property
def energy_kilojoule(self):
    if self.energy:
    ...
```

rolandgeider (changes requested):

rolandgeider:
While showingeverywhere both the kcal and the kJ would break the UI flow, I'd add them to the nutrition list and the nutritional data table under the plan, then this would be ready to merge

nopinter:
Hi, thanks for your feedback! We've implemented your suggested changes. We found the "4,184" / "4.184" issue very funny, but that now is fixed 👍 . Additionally, we wanted your feedback on an aesthetic issue, which we've attached a picture of. 

Screen Shot 2020-12-13 at 10 30 05 PM

As you can see in the red circle, sometimes the kilojoules wrap around onto the next line. Let us know your thoughts on this. We'd be more than willing to change some of the HTML/CSS to make it look cleaner if necessary.

rolandgeider:
I also had to laugh when I opened a plan and saw like half a million kJ 😂

And as for the line wrapping issue, we can save some space by removing the decimal place but other than that I wouldn't change much

rolandgeider (approved):

nopinter:
We have experimented with removing the decimal place, but we are still observing the line wrapping issue. We didn't make any further changes for this reason

nopinter:
Also, we would be willing to take #581 instead of #575 ! I'll will comment on those to reflect this

rolandgeider (approved):

rolandgeider:
Merged, thanks!!

Issue #551, opened 2020-11-17
Check name similarity when adding exercises


rolandgeider:
When adding exercises, we should check if there is already an existing one with a similar name, e.g. with the Levenshtein distance between it and the names already in the database. 

Ideally, we would show the possible matches and ask the user if he really wants to continue, but it's ok to just show an error for now.

TODO: update the exercise model's name_clean() and show an error message. We would need to experiment a bit on what distance is appropriate. Perhaps 4 oder so? Important: check only against other exercises with the same language

lydiaxing:
Hi, I'd like to work on this!

rolandgeider:
Hi there! It's all yours :)

PR #578, opened 2020-12-13
Check name similarity when adding exercises

lydiaxing:
### Proposed Changes
Addressed [issue #551](https://github.com/wger-project/wger/issues/551) by adding a cleaning function to the exercise submission form which checks for name similarity to existing exercises via the levenshtein algorithm
image

and isort (``isort``) 

### Other questions

this PR?)
No

Thank you!


wger/utils/language.py
```diff
@@ -99,7 +100,7 @@ def load_ingredient_languages(request):
     """
 
     language = load_language()
-    languages = load_item_languages(LanguageConfig.SHOW_ITEM_INGREDIENTS)
+    languages = load_item_languages('2')
```
rolandgeider, line 103: Mhh, having magic numbers (or strings) is really not a good idea. We could move SHOW_ITEM_INGREDIENTS to utils/constants.py, that would probably break the circular import. Having the import inside the function is OK if it's necessary, but better write a short comment why.

The whole LanguageConfig for ingredients will ideally go away when the Open Food Facts issue is done and the DB has more products (the config is used so that if e.g. a language has only a handful of exercises, you can configure the application to also show those users exercises in english and the application is more useful)

rolandgeider:
Thanks for the PR! I want to do some more testing before I run this on the production database (on a test run I noticed how there are almost-but-not-quite duplicates of products) and also need to think about how to distribute the fixtures after this

wger/exercises/views/exercises.py
```diff
@@ -157,6 +160,59 @@ def view(request, id, slug=None):
...
+    class Media:
+        js = (settings.STATIC_URL + 'yarn/tinymce/tinymce.min.js',)
+
+    def clean_name_original(self):
+        """
+        Throws a validation error if the submitted name is too similar to an existing
+        exercise's name
+        """
+        name_original = self.cleaned_data['name_original']
+        languages = load_item_languages(LanguageConfig.SHOW_ITEM_EXERCISES)
```
rolandgeider, line 202: You don't need to load the languages from config, here it's enough if you just filter by the one fom the current exercise. The LanguageConfig is meant so that if e.g. Russian has only a handful of exercises and ingredients, you can configure the application to also show exercises in English so that it's a bit more useful. In our case here, it's perfectly OK that there exist different exercises with the same name but in different languages

rolandgeider (changes requested):

rolandgeider (approved):

rolandgeider:
Merged, thanks!!

PR #579, opened 2020-12-13
Return muscle image URLs from model (fixes #547)

### LINKED ISSUE #547: Return image URL on muscle endpoint
The muscle endpoint should also return the URLs for the muscle images and the backgrounds (probably the exercise endpoint as well). Needs to check whether we can access the URL when the static files are hosted on AWS or such

gengkev:
### Proposed Changes

Each muscle is associated with a primary and secondary image, which are used if it is the primary or secondary muscle in an exercise, respectively. The previous code computed the URLs of these images in the `render_muscle()` function. This commit generalizes this slightly by moving that calculation to the muscle model instead.

Question: Does this address the intent of the original issue #547, or should additional refactoring be made? For instance, we could use an `ImageField`, similar to `ExerciseImage`, so that muscle images can be uploaded and edited by the admin (rather than only using the fixtures).

and isort (``isort``) 

### Other questions

* Does this PR introduce a breaking change such as a database migration? (i.e. what changes might users need to make in their running application due to this PR?)
  No

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,6 +38,7 @@ Developers
 * Hita Kambhamettu - https://github.com/Hita-K 
 * Derek Li - https://github.com/derekli17
 * Noah Pinter - https://github.com/nopinter
+* Kevin Geng - https://github.com/gengkev
 
 Translators
 -----------
--- a/wger/core/templates/tags/muscles.html
+++ b/wger/core/templates/tags/muscles.html
@@ -1,6 +1,6 @@
 {% load static %}
 {% if not empty %}
 <div class="muscle-background mx-auto"
-     style="background-image: {% for background in backgrounds %}url({% static background %}){% if not forloop.last %},{% endif %}{% endfor %};">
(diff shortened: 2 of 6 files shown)


rolandgeider:
This is fine, having an admin interface for the muscles would be too much work for something that is basically never used.

The only change I'd suggest is adding the muscle fields to MuscleSerializer (explicitly, instead of __all__, which is always sensible to do)

gengkev:
Thanks! Just to make sure, would the reason for that be to make sure that only the two actual properties (name and is_front) are serialized, while the url properties added by this change are not?

rolandgeider:
no, add the new url properties as well, it's useful having the background URLs when using the API

gengkev:
Thanks, that makes sense! I assumed that "serializer" meant the purpose was for storing data, but it makes more sense if the data is being used as an API instead.

I've changed the MuscleSerializer to do that, and tried to modify the `MuscleApiTestCase` to verify that the correct URLs are in fact being returned.

rolandgeider:
Looks good, thanks!!

gengkev:
I updated the PR to fix the merge conflicts!

rolandgeider:
Merged, thanks!

PR #584, opened 2020-12-15
Customizable unraid

aronwk-aaron:
### Proposed Changes
This is mainly a copy of the development docker extended to pull variables from the environment that can be set by docker
There may be a better way to do this rather than making a new docker folder, but I am not sure.

This allows for unraid (as well as other things too) to easily inject variables to the container.
Most things have a default or an enable flag, so things like email and redis don't have to be used, but can easily be configured

Unraid xml for community applications needs to be in a separate repo.
Here is a guide on what needs to be done to get added to CA
https://forums.unraid.net/topic/57181-docker-faq/#comment-566084

- [ NA] Tests for the changes have been added (for bug fixes / features)
- [ Y] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8``)
and isort (``isort``) 
- [ Y] Added yourself to AUTHORS.rst

### Other questions

this PR?)

No


aronwk-aaron:
Here is the associated unraid template.
https://github.com/aronwk-aaron/unraid-templates

aronwk-aaron:
Automated docker builds can be setup with dockerhub
ex: https://hub.docker.com/r/aronwk/wger
I just have it set to always tag latest and build off the branch I was working on, but this can be changed for cutting releases
image

rolandgeider:
Nice!! From a first glance, you didn't change anything on the dockerfile itself right? (besides the settings file) Being able to configure those via environmental variables would also be useful in the regular docker and docker-compose

extras/docker/unraid/settings.py
```diff
@@ -0,0 +1,117 @@
...
+print('''#############################################################
+    ENV VARS:''')
+print(os.environ)
+print('''#############################################################
+    WGER_SETTINGS''')
+print(WGER_SETTINGS)
+
+print('''#############################################################
+    Database:''')
+print(DATABASES)
```
rolandgeider, line 116: is it smart to print this? The DB password would be outputted as well

rolandgeider:
just have added a new repo with your unraid config

extras/docker/unraid/settings.py
aronwk-aaron: Good point, forgot that I added debug logging when I did this

aronwk-aaron:
> Nice!! From a first glance, you didn't change anything on the dockerfile itself right? (besides the settings file) Being able to configure those via environmental variables would also be useful in the regular docker and docker-compose

correct, I wasn't sure how useful this would be for other dockers, but as the default and debugging are still there, his could be rolled into the development docker.
With venv, you can set environmental variables to be exported when you activate, this is what I do on personal projects instead of having to set them every time.

Issue #580, opened 2020-12-13
Use explicit field lists in serializers or forms


rolandgeider:
Instead of using `__all__`, we should explicitly list all the model's list:

https://docs.djangoproject.com/en/3.1/topics/forms/modelforms/#selecting-the-fields-to-use

klu17:
I'll take this one @rolandgeider

rolandgeider:
All yours! :)

PR #588, opened 2020-12-21
Replace __all__ with explicit field lists in serializers and forms. #580

klu17:
### Proposed Changes

  - I replaced` __all__` with field lists that I created from information that I found in models.py files. Issue #580

and isort (``isort``) 

### Other questions

this PR?)
No.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ Developers
 * Derek Li - https://github.com/derekli17
 * Noah Pinter - https://github.com/nopinter
 * Kevin Geng - https://github.com/gengkev
+* Karmen Lu - https://github.com/klu17
 
 Translators
 -----------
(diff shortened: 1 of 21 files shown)


klu17:
I noticed that tests test editing and creating of instances of models like Gym. Should the tests pass as is? If so, I will remove extra fields from the field lists that I have added. ... I wasn't sure if I should update the tests because this is neither a bug fix or a new feature. Additionally, I am unsure of what the desired functionality is? 

For example, when creating a new Gym in the past, the test passes when a user provides a value for the name field. A Gym has other fields too like phone, email, owner, etc.

rolandgeider:
That was quick! :)

Good question. At least for the non-editable fields setting them to read-only would probably fix this (we might have to change that in the future for the flutter application). And the tests for the gym are alright the way they are, it's not necessary to test for every possible combination

klu17:
@rolandgeider Should be ready for review.

rolandgeider:
Thanks!

PR #592, opened 2020-12-29
Add VideoReference to models.py under exercise.

klu17:
### Proposed Changes
- Add a VideoReference model with URL, description, and ForeignKey to exercise. 

- [ ] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8``)
and isort (``isort``) 

### Other questions

this PR?)

### DIFF
--- a/wger/exercises/models.py
+++ b/wger/exercises/models.py
@@ -545,3 +545,26 @@ class ExerciseComment(models.Model):
         Comment has no owner information
         """
         return False
+
+
+class VideoReference(models.Model):
+    """
+    Model for a video reference that demonstrates an exercise.
+    """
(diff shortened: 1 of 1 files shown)


klu17:
Should I create a view or add tests as well? I referred to ExerciseComment in the model. I am unsure if there are methods/functions to add for this. @rolandgeider

rolandgeider:
Hi! Yes, we will also need to connect this to the view. We will need a form, validation so that e.g. only youtube URLs are accepted and then also embed the video itself

rolandgeider:
actually... since this is user submitted, we to check the submission status for this before displaying. But we can first implement it so that only an administrator is allowed to do this and then expand it

PR #611, opened 2021-02-28
Limit how often set rows can be updated to avoid excessive updates to the form

### LINKED ISSUE #610: "Number of sets" slider often adds the wrong number of rows to the form
### Summary
The "Number of sets" input on the "Add exercises" form is styled to look like a slider.  This encourages users to hold down the "handle" of the slider, and move it to the position they wish to set.

If the slider is used in this way, the number of sets that are added to the form becomes unpredictable.  E.g. setting to 9 sets, and then down to 1 set will often show 5 or 6 sets rather than 1.

### Recreation steps

**Given** I am logged in as a User
**And** I am viewing a workout /workout/{id}/view/

**When** I click "Add exercises to this workout day"
**And** I search for and add an Exercise, e.g. Barbell Ab Rollout
**And** I click and drag the "Number of sets" control to a higher value e.g. 8
**And** I click and drag the "Number of sets" control back to a lower value e.g. 2

**Then** it should always show the correct number of set rows.
**Instead** there is often more set rows than there should be. 

**Note**: you can see this consistently if you click the handle and just "shake" the value between 2 extremes.  The javascript struggles to keep up with the alterations it is being told to make to the page.

Occurs on Chrome 88 (Windows 10) and Firefox 85 (Windows 10).  Likely to occur on other browser/OS combinations too, these are just the two I have immediate access to.

If the video isn't working, use the seek bar to go a few seconds into it, and it seems to work after that.  Not sure why!

https://user-images.githubusercontent.com/38165285/109400107-a77f4080-793e-11eb-807d-ae0cfffffde3.mp4

qagoose:
### Proposed Changes

  - Debounce the sets slider on the add exercise form to prevent excessive updates to the form (fixes #610 )

and isort (``isort``) 

No tests added, as this is a JavaScript fix, and I don't believe there are any JS/UI tests.

Checked in recent versions of Chrome, Firefox, and Edge (Chromium) and all behave as expected.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -51,6 +51,7 @@ Developers
 * Aaron Kimbrell - https://github.com/aronwk-aaron
 * Aaron Bassett - https://github.com/aaronbassett
 * Rachael Wright-Munn - https://github.com/ChaelCodes
+* Will Pearson - https://github.com/qagoose
 
 
 Translators
--- a/wger/core/static/js/wger-core.js
+++ b/wger/core/static/js/wger-core.js
@@ -97,6 +97,17 @@ function wgerSetupSortable() {
   });
 }
 
+/* 
+Functions related to site performance 
+*/
+
+function debounce(func, delay, timer) {
+  // Delays calling a function until it hasn't been called for the set delay
+  clearTimeout(timer);
+  timer = setTimeout(func, delay);
+  return timer;
+}
+
 /*
  Functions related to the user's preferences
  */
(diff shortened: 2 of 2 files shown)


rolandgeider:
Nice! I'll test this later today and merge it.

rolandgeider:
I have tried this and the problem is now slightly different. For example, when you are on 4 sets and turn the slider to one, the end result is 3. I think sending the request on mouseup would be the better solution here 

image

qagoose:
I've updated this to use the 'pointerup' event now.  'mouseup' did not work on my Android Chrome 88, and I've seen contradictory statements about whether it works beyond Chrome 55.  My experience is it didn't work at all.

Pointer Event API seems to have fairly widespread support: https://caniuse.com/pointer

I don't think this fully fixes the issue, but for the majority of users it should be a better experience now.

rolandgeider:
Good call, this should make the dialog more usable till we come up with a better solution

Issue #585, opened 2020-12-16
Log in as user functionality produces 403 error 


Ownercz:
Hello,  
when trying to log in from administrator account as user **without** default gym set (that happens when you register through site yourself), I get an 403 error.  

#### Steps to reproduce
Use any of the current deployment methods (docker, docker-compose, VM installation with Debian 10).  And try as an admin to use **log in as this user** (e.g. http://url/cs/user/5/trainer-login ).
Test user is guaranteed to produce this error if no gym is set:  
![image](https://user-images.githubusercontent.com/1702837/102336781-ee8da480-3f91-11eb-81a7-14267e3cef13.png)
#### Log
```
[Wed Dec 16 11:09:28.687239 2020] [wsgi:error] [pid 11615:tid 140290494953216] [remote 192.168.4.14:60186] Forbidden: /cs/user/5/trainer-login
```

Ownercz:
With further testing this happens in two cases:

1. when user is either **not** in default gym 
2. or **does not have** any gym at all. 

So creating another gym and adding user there results in the same error.

PR #620, opened 2021-03-08
fix(): 404 if a trainer tries to login to a different gym

bradsk88:
Rather than just returning "permission denied", we can interpret this scenario
as a "failure to look up a user with that gym/user ID combo".

This should give users the information they need to self-recover, while not
leaking any other sensitive details.

Fixes: #585

### Proposed Changes

**Before**: Scary "permission denied" page
simplescreenrecorder-2021-03-07_21 00 45

**After**: Helpful 404 page with explanation
simplescreenrecorder-2021-03-07_20 45 28

and isort (``isort``) 

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -28,9 +28,9 @@ describe a use-case or similar.
 ## Pull Requests
 You want to contribute code? Awesome! Here are some tips:
 
-* Make sure the tests are running (``python3 manage.py tests``)...
+* Make sure the tests are running (``python3 manage.py test``)...
 * ... and if you write new code, write new tests
-* Lint the code with flake8 (``flake8 --config .github/linters/.flake8``)
+* Lint the code with flake8 (``flake8 --config .github/linters/.flake8 ./wger``)
 and isort (``isort``) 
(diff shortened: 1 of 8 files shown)


bradsk88:
I should note that I think the "real" fix is probably to grey out the "log in as user" button if there is a gym ID mismatch.

But, I figured it was a good practice to address this at the handler-level first.

bradsk88:
`test_logout` is failing, but I honestly can't tell what that test is supposed to be asserting.

https://github.com/bradsk88/wger/blob/505637579cc159c556518a8c1d11782d10f24142/wger/gym/tests/test_user.py#L239-L248

If someone can clarify, I'd be happy to try and fix it.

rolandgeider:
Nice! You are right that we should hide the button, but this is also a useful usability improvement.

The tests check that after logging in as user 2, the trainer identity is stored in the session and deleted after returning to the own account

bradsk88:
@rolandgeider Should be good to go.  Just had my mocks set up incorrectly.

rolandgeider:
Awesome, thanks!

PR #631, opened 2021-03-13
fix(): Disable "Log in as this user" when gym doesn't match

### LINKED ISSUE #585: Log in as user functionality produces 403 error

bradsk88:
I'm not the biggest fan of using native tooltips since they're so slow to render.
But I didn't see any other options being used in this project.

Deviated from bootstrap styles because I didn't feel their "disabled" state did
a very good job of communicating that the button is disabled.

I prefer leaving the button in place rather than hiding it so that users know
what is theoretically possible if they configure the system correctly.

- [x] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8 ./wger``)
and isort (``isort``) 

# Before 
Button is enabled but takes you to a 404
simplescreenrecorder-2021-03-07_20 45 28

# After
Button is disabled and shows a cursor/tooltip

https://user-images.githubusercontent.com/2147630/111019954-22f1e080-8388-11eb-9809-4185fca8d176.mp4

Fixes #585

### DIFF
--- a/wger/core/templates/user/overview.html
+++ b/wger/core/templates/user/overview.html
@@ -469,6 +469,26 @@
 {#         #}
 {% block options %}
 {% if perms.gym.gym_trainer %}
-    <a href="{% url 'core:user:trainer-login' current_user.pk %}" class="btn btn-success btn-sm">{% trans "Log in as this user" %}</a>
+
+    <style>
+        .btn.btn-sm.btn-secondary {
+            background: #EEE;
+            color: #AAA;
+            cursor: not-allowed;
+        }
+    </style>
+
+    <a
+       {% if enable_login_button %}
+        href="{% url 'core:user:trainer-login' current_user.pk %}"
+        class="btn btn-sm btn-success"
+       {% else %}
+        href="#"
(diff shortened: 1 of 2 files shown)


bradsk88:
Just noticed if you try to "log in as yourself" as an admin, you get a 403. Happy to make that better in another PR.

rolandgeider:
> Just noticed if you try to "log in as yourself" as an admin, you get a 403. Happy to make that better in another PR.

🙏🏻

rolandgeider:
Thanks for the PR

Issue #599, opened 2021-01-24
Add UI for uploading exercise images


12people:
Looking at the Wger web interface at wger.de, there is a way to add and edit exercises, but not a way for users to upload images corresponding to those exercises.

It'd be great if users could upload images right from the web GUI.

rolandgeider:
Definitely. This and the other colaboration focused issues will be (one of) the focus for the next release

sophialr:
working on this with @emuth

rolandgeider:
To clarify, there is currently a UI to upload images to exercises, but only for administrators. Regular users should be able to do this as well. To avoid spam, there needs to be a process before the images go "live". At the moment an administrator needs to manually OK every submitted exercise and does not scale. We could add a flag to the user profile to indicate whether the user is more trustworthy, e.g. after successfully having submitted an exercise or if the account is older than X days

emmaamuth:
@rolandgeider thank you for the clarifications! So are you thinking then if the user isn't "trustworthy" yet then that process of an admin manually OK-ing the image should still apply? Any preference on how many days before an account is considered "old enough"?

rolandgeider:
yes exactly.

And for the preferences, I would say 3 weeks? In order to allow for more flexibility we could add a property to the preferences model, that way we can add other checks in the future (such as "already submitted an exercise"). I'm thinking about something along the lines of


```python
@property
def is_trustworthy(self):
    joined_days_ago = datetime.today() - self.user.date_joined
    return joined_days_ago > 21
```

emmaamuth:
@rolandgeider perfect, that all makes sense. Thank you again!

PR #662, opened 2021-04-23
Added UI and credibility checks for users to upload exercise images

emmaamuth:
### Proposed Changes

  - Addresses Issue #599 
  - Now users who fulfill a trustworthiness criteria (has flexibility for additional criteria to be added) can add images to exercises without waiting for admin approval
  - In order to make the change above, updated several permissions checks to depend on the group a user is a part of instead of them having one or more exercise permissions because without that change a trustworthy user could view UI for all admin features
  - Also resolved small typo and indentation error

- [X] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8 ./wger``)
and isort (``isort``)

### Other questions

this PR?) No

### DIFF
--- a/wger/core/models.py
+++ b/wger/core/models.py
@@ -21,12 +21,12 @@ import decimal
 # Django
 from django.contrib.auth.models import (
     Permission,
-    User
+    User,
 )
 from django.core.exceptions import ValidationError
 from django.core.validators import (
     MaxValueValidator,
-    MinValueValidator
+    MinValueValidator,
 )
 from django.db import models
 from django.db.models import IntegerField
@@ -224,28 +224,27 @@ by the US Department of Agriculture. It is extremely complete, with around
     @property
     def is_trustworthy(self):
         """
-        Trustworthy if they've had an account for more than three weeks. More trustworthiness
-        criteria could be added in future.
+        Returns true for trustworthy if they've had an account for more than three weeks.
+        More trustworthiness criteria could be added in future.
         """
-        days_since_joined = datetime.date.today() - self.user.date_joined.date()
 
-        if days_since_joined.days > 21:
+        days_since_joined = datetime.date.today() - self.user.date_joined.date()
+        days_in_three_weeks = 21
+
+        if days_since_joined.days > days_in_three_weeks:
             # perms will be updated only once, after they hit three week mark
             image_perm = Permission.objects.get(codename='add_exerciseimage')
             self.user.user_permissions.add(image_perm)
 
-        return days_since_joined.days > 21
+        return days_since_joined.days > days_in_three_weeks
 
     @property
     def has_exercise_permission(self):
-        """
-        Returns true if user has all the exercise permissions and not just one.
-        """
-        if self.user.groups.filter(name='admin').exists() or \
-           self.user.groups.filter(name='exercises_editor').exists():
+        """Returns true if user has all the exercise permissions and not just one."""
+        if self.user.groups.filter(name='admin').exists() or self.user.groups.filter(
+                name='exercises_editor').exists():
             return True
-        else:
-            return False
+        return False
 
     #
     # User statistics
--- a/wger/core/models.py
+++ b/wger/core/models.py
@@ -19,7 +19,10 @@ import datetime
 import decimal
 
 # Django
-from django.contrib.auth.models import User
+from django.contrib.auth.models import (
+    Permission,
+    User
+)
 from django.core.exceptions import ValidationError
 from django.core.validators import (
     MaxValueValidator,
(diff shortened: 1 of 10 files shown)


rolandgeider:
Thanks! I'll take a look tomorrow or the day after

wger/nutrition/templates/ingredient/form.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n %}
 
-{% if not perms.exercises %}
```
rolandgeider, line 3: This should actually have been `perms.nutrition.add_ingredient`

wger/core/models.py
```diff
@@ -218,6 +221,31 @@ class UserProfile(models.Model):
...
+            # perms will be updated only once, after they hit three week mark
+            image_perm = Permission.objects.get(codename='add_exerciseimage')
+            self.user.user_permissions.add(image_perm)
+
+        return days_since_joined.days > days_in_three_weeks
+
+    @property
+    def has_exercise_permission(self):
+        """Returns true if user has all the exercise permissions and not just one."""
+        if self.user.groups.filter(name='admin').exists() or self.user.groups.filter(
```
rolandgeider, line 244: I'm not sure if checking for the group is a good idea. While we don't do now, if we add in the future a new group that should also be able to edit exercises this would not work. Better to check for individual permissions

wger/exercises/templates/exercise/view.html
```diff
@@ -92,7 +92,7 @@ <h5 class="mt-4">{% translate "Description" %}:</h5>
 
 {# Images #}
 {% with images=exercise.images.accepted %}
-{% if images or perms.exercises.change_exerciseimage  %}
+{% if images or perms.exercises.change_exerciseimage or user.userprofile.is_trustworthy %}
```
rolandgeider, line 95: We should not allow users to edit or delete images 😅😅, just to submit them!

wger/core/models.py
```diff
@@ -218,6 +221,31 @@ class UserProfile(models.Model):
...
+        More trustworthiness criteria could be added in future.
+        """
+
+        days_since_joined = datetime.date.today() - self.user.date_joined.date()
+        days_in_three_weeks = 21
+
+        if days_since_joined.days > days_in_three_weeks:
+            # perms will be updated only once, after they hit three week mark
+            image_perm = Permission.objects.get(codename='add_exerciseimage')
+            self.user.user_permissions.add(image_perm)
```
rolandgeider, line 237: I'd rather not give the user the add_exerciseimage permission just yet, we can simply do the dynamic checks

wger/nutrition/templates/ingredient/form.html
emmaamuth: Gotcha, since we changed the permissions of a "normal" user we were updating all the places that were dependent on the old exercise check and this was one of them, didn't realize it wasn't supposed to be perms.exercises in the first place!

wger/core/models.py
emmaamuth: Makes sense! I believe if the strategy gets changed to doing the permission check dynamically as you mentioned below this will be irrelevant anyway.

wger/exercises/templates/exercise/view.html
emmaamuth: We had to add this check because when we were testing if they were just a normal user and there weren't already images, it never entered this conditional and they weren't able to see the "Add new image" link. When we tested it, they weren't able to edit or delete images because of the nested check on line 107. Does that seem correct?

wger/core/models.py
emmaamuth: Understandable! That was our original thought but we struggled to find where to do the dynamic check that would allow them to actually see the image adding form, so I guess do you have any guidance on where that would be? Originally we thought it was in wger/exercises/views/images.py class ExerciseImageAddView but couldn't see how to incorporate it without adding permissions.

wger/exercises/templates/exercise/view.html
rolandgeider: ahh, I see what you mean. I think that since all links are protected with corresponding permission checks, we could simplify this to {% if images %}

wger/core/models.py
rolandgeider: yeah, looking at the code it's a bit trickier that what I thought. One possibility would be to manually check for permissions or the trustworthy attribute by overwriting the dispatch method (you can then remove the PermissionRequiredMixin mixin and the permission_required). Something like this is done elsewhere, e.g. in the WgerPermissionMixin or WgerFormMixin where it checks the ownership of database objects.

rolandgeider:
I'm merging this (manually) into another branch with other exercise workflow improvements.

Thanks for the code!

Issue #656, opened 2021-04-13
Change naming logic for exercise image filename


rolandgeider:
## Use case

Exercise images should get a UUID as the name

## Proposal

At the moment, after adding an image, the file on the disk is named after the exercise, we should use the image's UUID for this.  See exercise_image_upload_dir in exercise/models

jackmulligan-ire:
I'm happy to do this!

jackmulligan-ire:
I'd be happy to do this.

rolandgeider:
It's yours

jackmulligan-ire:
Submitted PR there.

PR #675, opened 2021-05-08
UUID rename

jackmulligan-ire:
### Proposed Changes

  - PR for (https://github.com/wger-project/wger/issues/656)  - Changed naming logic for exercise image filename 

- [x] New python code has been linted with with flake8 (``flake8 --config .github/linters/.flake8 ./wger``)
and isort (``isort``) 

### Other questions

this PR?) - No.

### DIFF
--- a/wger/exercises/models.py
+++ b/wger/exercises/models.py
@@ -18,6 +18,7 @@
 # Standard Library
 import logging
 import uuid
+import pathlib
 
 # Django
 from django.conf import settings
@@ -439,7 +440,7 @@ def exercise_image_upload_dir(instance, filename):
     """
     Returns the upload target for exercise images
     """
-    return "exercise-images/{0}/{1}".format(instance.uuid, filename)
+    return "exercise-images/{0}/{1}{2}".format(instance.exercise_base.id, instance.uuid, pathlib.Path(filename).suffix)
 
 
 class ExerciseImage(AbstractSubmissionModel, AbstractLicenseModel, models.Model):
--- a/wger/exercises/models.py
+++ b/wger/exercises/models.py
@@ -440,7 +440,8 @@ def exercise_image_upload_dir(instance, filename):
     """
     Returns the upload target for exercise images
     """
-    return "exercise-images/{0}/{1}{2}".format(instance.exercise_base.id, instance.uuid, pathlib.Path(filename).suffix)
+    ext = pathlib.Path(filename).suffix
+    return "exercise-images/{0}/{1}{2}".format(instance.exercise_base.id, instance.uuid, ext)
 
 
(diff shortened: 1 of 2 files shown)


wger/exercises/models.py
```diff
@@ -439,7 +439,7 @@ def exercise_image_upload_dir(instance, filename):
     """
     Returns the upload target for exercise images
     """
-    return "exercise-images/{0}/{1}".format(instance.exercise_base.id, filename)
+    return "exercise-images/{0}/{1}".format(instance.uuid, filename)
```
rolandgeider, line 442: This only changes the subfolder :)

Instead of  e.g. ``exercise-images/5/the_original_filename.jpg``, it should be ``exercise-images/5/abcd....jpg``. With the current code it would result in ``exercise-images/abcd..../the_original_filename.jpg``

I was also thinking that it would be useful to preserve the file extension, you can easily extract it with ``pathlib.Path(filename).suffix``. Just add it as a third parameter of the format call (and in the string like ``...{0}/{1}.{2}``)

rolandgeider (changes requested):

wger/exercises/models.py
jackmulligan-ire: Hey, ah right I get you, will change it around and submit a new PR.

Also not sure why some of the CI actions failed, any ideas?

rolandgeider: No problems!  You can just push a new commit to the same branch, the PR will update.

And don't worry about the CI, I haven't updated the main server yet and it's trying to download some images to build the docker file

jackmulligan-ire: Grand, so pushed new commit there. I wrote it in the format "exercise-images/{0}/{1}{2}", I was getting "..jpg" instead of ".jpg" at the end if I took "exercise-images/{0}/{1}.{2}".

Ran the tests and it failed 1 test in lines 208-210 in base_testcase.py. Test is related to image names, so assuming it failed because of the new format and all is well. I took a look in base_testcase.py to see if I could rewrite the test for the new format but it would be a bit beyond my skills at the moment, still getting the hang of OOP!

rolandgeider: You are right, those compares the image files. I'll take a look later (that part of the tests are really hairy, usually they are not like that)

rolandgeider:
it seems the linter is failing, the line with the suffix is longer than 100 characters. If you fix that I can merge this

jackmulligan-ire:
Should be good to go now, linter ran clean for me in terminal.

rolandgeider (approved):

rolandgeider:
Thanks, merged

This was your first or right? In that case congratulations!

Issue #773, opened 2021-08-22
Allow user to give meals a description


rolandgeider:
## Use case

Give meals a textual description / name such as "Breakfast" or "after workout"

## Proposal

The meal model should get a new optional field "name", 25 characters long sounds enough. This field also needs to be exposed over the API

PR #774, opened 2021-08-22
Allow user to give meals a description 

Gr8ayu:
Allow user to give meals a description/name such as breakfast or after workout
resolving #773 

# Proposed Changes

- add name field to meal model class 
- add 'name' in MealSerializer
- add 'name' in MealInfoSerializer

### Other questions

  (e.g. database migration)?
YES


rolandgeider:
That was quick 😄

Can you commit the migration file as well? Also, we should show the meal name in the template

Gr8ayu:
yeah I'll add those migrations, and by template you mean frontend right ?

rolandgeider:
Yes

rolandgeider:
And the last thing we need to do: add the new field to the list of fields in the forms in `MealCreateView` and `MealEditView`, otherwise it would not be possible to set it

Gr8ayu:
image

I've added meal name like this in the template

Gr8ayu:
> And the last thing we need to do: add the new field to the list of fields in the forms in `MealCreateView` and `MealEditView`, otherwise it would not be possible to set it

Okay

rolandgeider:
Great, thanks!

Issue #653, opened 2021-04-13
Add images to ingredients


rolandgeider:
Add images to ingredients

rolandgeider:
Implementing images has been started in this branch: https://github.com/wger-project/wger/tree/feature/ingredient-images

rolandgeider:
There is already working code that fetches the images from Open Food Facts and saves them to the local database (`get_image` in the ingredient model). This is fine for nutritional plans or such since after the first fetch the images are known locally, but when using the autocompleter this process can take a lot of time since we need to perform two request per ingredient. What needs to be done:

- return the local image for the autocompleter
- add a management command that iterates over all ingredients that are currently used in a nutritional plan and accesses the get_image property so that those are downloaded initially

rolandgeider:
Hi! Sure, it's yours. Yes, exactly, we can access the image in the model and return that if it exists

rolandgeider:
Backend is done, we just need to show the images in the autocompleter

ImTheTom:
I can see if I can the autocompleter soon.

ImTheTom:
Unless you wanted to finish it off @olijzenga

rolandgeider:
Note that this is not merged yet because we still need to think about performance issues.

Ideally we would download all images that have been used somewhere, so e.g. ingredients that are used in meals or logged to the diary, for this we will need to make sure this also works when using the rest api, since more and more parts of the application will be using it. I would also prefer to use something like celery (which we will need anyway if we implement data synchronization for other fitness services) so that this can happen in the background

aryandgandhi:
Hey @rolandgeider I'd like to work on this if it's still open, also reading through some of the comments, I'd like to implement a button below [no workouts found] which some pre-made workouts. Let me know what issues are good for first timers and that you still need help with.

rolandgeider:
hi @aryandgandhi ! This issue is almost done, just need to do some polishing and tweak a bit how we store license information. 

Any of the issues in the milestone [for version 2.2](https://github.com/wger-project/wger/milestone/11) would be very useful. Perhaps something like 817, 1003 or 1012?

PR #1001, opened 2022-04-05
Feature/ingredient images

olijzenga:
Based on #653 

# Proposed Changes

- Added ingredient image url to the search response in the same way that it is done for exercises.
- Added a script for downloading images from OFF for ingredients which are used in nutrition plans.

- (Only updated the test for the case that the ingredient does not have an image.)

### Other questions

  (e.g. database migration)?
No


olijzenga:
Note: setting up the database schema on this branch using a clean database gave me a migration conflict error. I worked around it by making a new migration using `manage.py makemigration --migrate`. However I did not include this migration in the pull request because I thought I might be doing something wrong (not much Django experience here). Let me know if I should include it.

rolandgeider:
You are right, there is a migration missing, feel free to commit it

olijzenga:
Done

wger/nutrition/api/views.py
```diff
@@ -167,11 +169,22 @@ def search(request):
         )
 
         for ingredient in ingredients:
+            if hasattr(ingredient, 'image'):
```
rolandgeider, line 172: can't we just do `if ingredient.image:` here?

olijzenga: I tried this but in that case RelatedObjectDoesNotExist is raised when the image is not there. The recommended way by Django to work around this is by using hasattr (according to https://docs.djangoproject.com/en/4.0/topics/db/examples/one_to_one/).

rolandgeider: you are right

rolandgeider:
Thanks for the PR! still need to do some things before merging to master, but I think the backend is done

Issue #984, opened 2022-03-05
Add categories to ingredients


rolandgeider:
## Use case

Add categories to ingredients

## Proposal

Improve the import script for the open food facts DB and import categories for ingredients

PR #1021, opened 2022-04-22
Proposed patch for Issue #984

HenleyChase:
# Patch for issue 984 "Add categories to ingredients"

- Used product.get method to get 'categories' and added
returned result to ingredient_data.
- Added myself to AUTHORS.rst

### Other questions

* I am unsure if result from product.get('categories', None)
is desired result. Please respond with suggestions on what
to change in order to get desired result from open food
facts database.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -64,6 +64,7 @@ Developers
 * absolutely-not-bot - https://github.com/absolutely-not-bot
 * Jaspreet Dhillon - https://github.com/jaspreetsd902
 * Sofiko Alaverdashvili - https://github.com/sophiamartelli
+* Chase Henley - https://github.com/HenleyChase
 
 Translators
 -----------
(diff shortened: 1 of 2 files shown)


rolandgeider:
Hi, thanks for the PR!

I think `product.get('categories', None)` is probably not going to be enough. If you look at the results from e.g. [this product](https://world.openfoodfacts.org/api/v0/product/3017620422003.json) there are several categories (which would also mean that we'll need to change the models). But more importantly, some of the categories are in English, some are translated ("en:breakfasts","en:sweet-spreads","fr:pates-a-tartiner") and we need to think on how to handle this. We could either just use all of them and simply add a language to the catetory so that we can filter later, only import the ones in English (will we have some incomplete products because of this? Do all products have at last all categories in English?) or something else. If we go the first route, we shouldn't create a new category each time, we would ideally just reference "en:breakfasts" so we could filter later. Most of this is not complicated, but one problem is that I don't know enough on how Open Food Facts handles the categories

PR #1032, opened 2022-05-05
Add result page for bmi calculator & unit (kg) for Weight

sunx14:
# Proposed Changes

- Add a fully functioning result page for the BMI calculator under nutrition dropdown menu so that it no longer returns just a simple HttpResposne.
- The result page shows the BMI result and also a pie chart that shows the percentage of people with different BMI in the US.
- Add the unit for Weight in the BMI calculator form to reduce confusion.

### DIFF
--- a/wger/nutrition/forms.py
+++ b/wger/nutrition/forms.py
@@ -101,6 +101,7 @@ class BmiForm(forms.ModelForm):
     weight = forms.DecimalField(
         widget=Html5NumberInput(),
         max_value=999,
+        label=_('Weight (kg)'),
     )
 
     class Meta:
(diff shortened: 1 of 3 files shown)


rolandgeider:
Thanks for the PR! The JSON response was being shown because we were using a deprecated d3 function, I've fixed that in https://github.com/wger-project/wger/commit/b8d7d33b38c2ced51c7e7903d2c8e687b4d2e053 so that the chart at least updates and shows the user's BMI. What's missing is showing the different scales for the different BMIs, I'll try to look another day but since we're planing on moving some of the frontent to react to have a clean architecture, I'm not sure how much time we should invest in that

PR #1038, opened 2022-05-06
Add correct units for bmi calculator and weight entries 

jiany18:
# Proposed Changes

-  When I was using the web app, I was really confused about which units the BMI calculator is using for weights
- Added units to weight in BMI calculator and weight entry to avoid future confusion


rolandgeider:
Thanks for the PR! There are actually a couple of places where we do consider whether the user uses metric or imperial units, but yeah, it would be better to explicitly display the units and change them according to the profile settings. The BMI calculator actually is aware of this and will convert the values accordingly in [calculate_bmi()](https://github.com/wger-project/wger/blob/master/wger/core/models/profile.py#L408)

PR #1040, opened 2022-05-11
Changes to prettify autocomplete and found an error in demo

syhong2:
# Proposed Changes

- Created an indent for suggestion category
- Pushed suggestion items a litter further to emphasize hierarchical structure of autocomplete suggestsions
- Found an issue (on the live site) where if you select an exercise from the autocomplete > hit the back button > try to click the search box again, it takes you straight to the exercise page again. Attempted to fix, but personally do not have enough web dev background to take up the task.
- Found an inconsistency between number of min characters for autocomplete in exercises by category and by equipment.

### Other questions

  (e.g. database migration)?
-No

### DIFF
--- a/wger/core/static/css/workout-manager.css
+++ b/wger/core/static/css/workout-manager.css
@@ -161,6 +161,7 @@ tbody tr:hover td .editoptions {
 .no-hover a:hover,
 .no-hover:hover {
   background-color: transparent;
+  cursor: pointer;
 }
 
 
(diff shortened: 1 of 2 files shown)


rolandgeider:
Thanks for the PR! I'll see that I fix that inconsistency.

There are a lot of shortcomings from the current solution (or the stone age mess of JS code that we still use), that's why we're reworking some of the UI with react, so I wouldn't invest too much effort in it, it just has to keep on working till it it's replaced, whenever that might be 😅

PR #1041, opened 2022-05-11
Draft: Improve muscle recognition

ImTheTom:
# Proposed Changes

- Adds a name_en column to the muscle view. This can be used as an alternative name.

- [X] Tests for the changes have been added (for bug fixes / features) (Very minimal though. Might make some more)

### Other questions

  (e.g. database migration)? Yes. Added a migration in exercises.

* Placing it up as a draft at the moment for general feedback on what I'm missing.

### Images
Muscle-alt-name
### Json response
```
        {
            "id": 11,
            "name": "Biceps femoris",
            "is_front": false,
            "name_en": "Hamstrings",
            "image_url_main": "/static/images/muscles/main/muscle-11.svg",
            "image_url_secondary": "/static/images/muscles/secondary/muscle-11.svg"
        },
        {
            "id": 13,
            "name": "Brachialis",
            "is_front": true,
            "name_en": "",
            "image_url_main": "/static/images/muscles/main/muscle-13.svg",
            "image_url_secondary": "/static/images/muscles/secondary/muscle-13.svg"
        }
```


rolandgeider:
Looks great!

What still needs to be done is mark those new names as translatable so they can be imported into weblate. That's the `extract-i18n.py` file I was talking about in the issue. Basically just adding a `[i.name_en for i in Muscle.objects.all() if i.name_en]` to the list of items should be enough. Then we can also do `{% translate muscle.name_en %}` in the template

I'm also thinking about directly updating the names in the migration file, then there's no need to run an additional step to load the fixture. Something like along these lines:

```python
muscle_names = [
    (1, 'Biceps'),
    (2, 'Shoulders'),
    (4, 'Chest'),
    (5, 'Triceps'),
    ...
]


def set_names(apps, schema_editor):
    Muscle = apps.get_model('exercises', 'Muscle')
    for mapping in muscle_names:
        m = Muscle.objects.get(pk=mapping[0])
        m.name_en = mapping[1]
        m.save()

 operations = [
        ....
        migrations.RunPython(set_names),
    ]


```

ImTheTom:
Yeah sweet. Thanks for the feedback. I'll look into it tonight 👍

wger/core/management/commands/extract-i18n.py
```diff
@@ -46,6 +46,7 @@ def cleanup_name(name: str) -> str:
         data = [i for i in ExerciseCategory.objects.all()] \
             + [i for i in Equipment.objects.all()] \
             + [i for i in Muscle.objects.all()] \
```
rolandgeider, line 48: we can remove this line :)

rolandgeider: Nevermind, could remove this by fixing the merge conflicts

ImTheTom: Should I regen the translation files I saw `wger/locale/el/LC_MESSAGES/django.mo` get changed when running the commands I found on https://wger.readthedocs.io/en/latest/i18n.html

rolandgeider: sure, go ahead

rolandgeider:
🚀

PR #1082, opened 2022-06-30
Draft: Show indepth exercise history changes

ImTheTom:
# Proposed Changes
- Adds a new page to show in depth exercise cahnges

- [ ] Tests for the changes have been added (for bug fixes / features) TODO

### Other questions

* Hey just placing this up to show what I've got if you have any concerns. At the moment the activty streams data column holds the associated history ID.
* Still left to do is to look into showing off create and delete changes and handling their reverts.
* Also need to make the code more defensive / write tests.
* Might also look into using the revert_url on history objects. 
When trying to access the revert_url on the history object at the moment it displays
```
LookupError: No installed app with label 'admin'.
```

Here's a video of it currently.
Shows updating exercise object id 1 author and description and then only reverting the license author.

https://user-images.githubusercontent.com/24668318/176581225-12b8f9a3-7aa6-4cc0-8879-3e25a49bd847.mp4


ImTheTom:
Yeah all good. Have fun. :)

ImTheTom:
Hey, was planning on adding support for created, updated and deleted exercises on the this page an option to revert each action.

Just realised when you delete a model it also deletes all the actstream actions for the model. Looking at their closed issues it seems like they don't want to support an option to keep them when you delete the model as well. So, I would have to swap around the logic to have the history models know about the activity streams.

Just checking with you if you think having the option to revert deletions is worth it as well, guess we wouldn't know who deleted it as the activity streams would be gone, but having an option to add it back would be nice.

rolandgeider:
mhh

I would say it is OK if we lose the information when an exercise is deleted. This will be only done by admins and we don't need to document everything, we're not a bank

ImTheTom:
Yeah makes sense, sounds good. I'll clean it up and make sure everything works tomorrow.

rolandgeider:
I simplified the logic here so that we don't need to explicitly save the history ID. I'm trying to finish everything left on the crowdsourcing branch so we can finally finally realease that stuff. In any case thanks again for the PR

PR #1142, opened 2022-10-08
Draft: #1137 Collect all authors related to ExerciseBase

### LINKED ISSUE #1137: Collect all authors for exercise base
At the moment the different parts of an exercise have separate author histories (base, translation, images, videos, comments) thorugh the magic of `AbstractHistoryMixin`. We should collect all these separate infos for the base and expose it as a separate property and endpoint since for the end user there only the exercise. While this is an expensive operation DB wise, the results are cached anyway so this should be OK

AdamPetik:
# Proposed Changes
resolves #1137 
-
-
-

### Other questions

  (e.g. database migration)?

### DIFF
--- a/wger/exercises/api/serializers.py
+++ b/wger/exercises/api/serializers.py
@@ -369,4 +369,5 @@ class ExerciseBaseInfoSerializer(serializers.ModelSerializer):
             "images",
             "videos",
             "author_history",
+            "total_authors_history",
         ]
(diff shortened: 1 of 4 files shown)


AdamPetik:
@rolandgeider I updated the AbstractHistoryMixin to recursively collect authors history for specified attributes in COLLECT_AUTHORS_FOR of particular model. What do you think of such approach? If it's unnecessary over complicated I can just collect authors for this issue specific case. I just thought it could be useful to generalize it a bit for the future usage.

rolandgeider:
I think a simple solution with only the handful of models is enough, we probably won't be adding any new ones for the forseable future

rolandgeider:
Yeah, I think this is better

PR #1161, opened 2022-10-22
Alternate rendering in case of an invalid password reset link

### LINKED ISSUE #1154: Not possible to reset password
```
AttributeError at /en/user/password/reset/check/abc/set-password
'NoneType' object has no attribute 'is_bound'
```

RohanKaran:
# Proposed Changes
- #1154 is occuring when uidb64 or the token is invalid in password reset link
- Added Alternate rendering in case of an invalid url to catch the error
- Fixes #1154

### DIFF
--- a/wger/core/templates/form.html
+++ b/wger/core/templates/form.html
@@ -15,7 +15,11 @@
 {% block title %}{{title}}{% endblock %}
 
 {% block content %}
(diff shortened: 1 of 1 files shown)


rolandgeider:
Thanks, will test this tomorrow!

PR #1167, opened 2022-10-28
added new code to features.html, css file and new images in static fo…

saket2001:
Worked on rework of landing page issue. Designed the new landing page using html bootstrap and css.


rolandgeider (approved):

rolandgeider:
Note to me: PR is approved but should be merged only when we upgrade bootstrap everywhere else as well

https://github.com/wger-project/wger/issues/1109

Issue #1124, opened 2022-09-20
BMI calculator only takes cm for height even when Imperial units are selected in preferences


deddc23efb:
## Steps to Reproduce

1. ... Current version 2.1a3.dev0 docker
2. ... enter the nutrition BMI calculator
3. ...

**Expected results:** I expect the data entry in the BMI calculator to honour the units I select in the user preferences.

**Actual results:** The units are set to cm and dont allow imperial units to be used in the BMI calculation.  

</details>

bernardokoen:
Hello, would like to claim this issue. Thinking of having the input field take in feet and inches when the user has selected imperial units as its preference and then converting that to cm on the backend and making the same function calls that are currently made for cm in order to keep it consistent. 

Please let me know your thoughts on this.

@deddc23efb @rolandgeider

rolandgeider:
Hi @bernardokoen 

Yeah, in general support for imperial units could be improved. Here we could show change the input based on the user preferences, always saving to cm seems like a good idea.

Additionally, this seems like a good opportunity to implement the page in react

PR #1185, opened 2022-11-17
[Issue 1124] Adds full imperial unit functionality to BMI Calculator

bernardokoen:
# Proposed Changes
Addresses https://github.com/wger-project/wger/issues/1124.
1. Displays appropriate units on BMI calculator for user based on unit preference.
2. Converts height and weight from imperial units to metric units when calculating BMI with imperial units selected.
3. Displays proper unit ranges in an error message when the user submits an inappropriate (out-of-range) height. Also gives a 406 error in this case.
4. Adds tests to ensure correctness of the above functionality.

### Other questions

  (e.g. database migration)?
No

Changes made in collaboration with @gabeliss


bernardokoen:
@rolandgeider Would love for you to take a look and let me know what you think when you get a chance. We were able to add these changes in to the existing framework without requiring much refactoring or restructuring.

rolandgeider:
nice!

I'll try to take a closer look today or tomorrow, but this looks good. The only thing I'd ask you is to mark the error messages and other user facing strings as translatable

bernardokoen:
Thanks! Please let me know if you have any other questions/concerns. 

When you say mark the strings as translatable do you mean within the Pull Request or within the source code? Do you have an example we could reference? I dug around and was unable to find one.

rolandgeider:
Sure, in nutrition/forms.py, the strings that are marked with `_("foo")`, then they can be later extracted and go to weblate so that the translators can translate it

bernardokoen:
Thanks for sharing that. I've applied those changes to the PR. Let me know if that doesn't look right or anything else needs to be tweaked! Thanks for your help.

wger/nutrition/views/bmi.py
```diff
@@ -69,12 +81,24 @@ def calculate(request):
...
             'weight': form.cleaned_data['weight'],
-            'height': request.user.userprofile.height,
+            'height': output_height,
         }
         data = json.dumps(result, cls=helpers.DecimalJsonEncoder)
-
+        response = HttpResponse(data, 'application/json')
+    else:
+        help_message = {
+            _('error'): _('Please make sure your height is within the appropriate range.'),
```
rolandgeider, line 90: don't need to translate the dictionary key :)

rolandgeider:
Looks good, thanks! Just added a small comment, then we should be able to merge.

Right now you have to enter the height in inches which I'm guessing is probably not very usual? On the other hand adding a real parser to be able to read/write from/to 5'10" is probably a pain, so this can definitely stay

bernardokoen:
We definitely agree–we considered using a foot and inches format but we thought it would add some unnecessary complexity and thought it would be best to keep in the same format as the cm. Took care of the remaining issue as well.

bernardokoen:
@rolandgeider we're seeing some issues with the _ before the translatable strings not being recognized. Have you run into this before?

bernardokoen:
Ah got it, thanks for clarifying. Just threw that in there.

PR #1188, opened 2022-11-19
Add MINIMUM_AGE_TO_TRUST variable

### LINKED ISSUE #1187: Remove or edit minimum account age for trustworthy
## Use case

There is a hardcoded [21 days minimum account age limit ](https://github.com/wger-project/wger/blob/da3a478c2caf5e8204d1f3322f08c8cbdc94be18/wger/core/models/profile.py#L204) That I think there should be a way to modify it using configs. Actually it doesn't make sense to prevent users from contributing in exercises in a private instance where all users are trusted.

## Proposal

There should be some config to modify it dynamically.

mohammadrafigh:
Closes #1187 

# Proposed Changes

- Added an environment variable to set minimum account age to trust.

I'm not a Python developer, so please review carefully.


rolandgeider:
Yes this is a good idea. I'm not sure right now if we mention the specific account age somewhere in the UI or just say something more general like the account needs to be of a certain age.

rolandgeider:
And hooray for tests! ;)

mohammadrafigh:
hmmm, yes it seems it's set in constants in the react project https://github.com/wger-project/react/blob/b038be369dab81782fa2b612cf40efdd7f112a12/src/utils/consts.ts#L3

Sent a PR there too.

rolandgeider:
Looks good, thanks! And you even got to do a multi repo PR 😁

But now that I'm looking at it, perhaps we could name this something like `MINIMUM_ACCOUNT_AGE` or better `MIN_ACCOUNT_AGE_TO_TRUST`?

mohammadrafigh:
Yep. I think so. Let me rename it.

rolandgeider:
Merged, thanks

PR #1201, opened 2022-12-13
Make sure entrypoint is chown'ed as well

stavros-k:
Was looking something unrelated, and noticed this. With the `.` you only chown the `WORKDIR`

# Proposed Changes

-
-
-

### Other questions

  (e.g. database migration)?

### DIFF
--- a/extras/docker/development/Dockerfile
+++ b/extras/docker/development/Dockerfile
@@ -50,7 +50,7 @@ COPY ${DOCKER_DIR}/entrypoint.sh /home/wger/entrypoint.sh
 RUN chmod +x /home/wger/entrypoint.sh
 RUN pip3 install --no-cache /wheels/*
 
-RUN chown -R wger:wger .
+RUN chown -R wger:wger /home/wger
 
 USER wger
 RUN mkdir ~/media \


rolandgeider:
Thanks for the PR.

I'm not sure if there was a reason why we only chown the source folder. Alternatively, couldn't we just update the chmod line above?

stavros-k:
The `chmod` does not do `chown` actions. Or am I missreading something?

rolandgeider:
no, but I meant that we could change the `chmod +x` line to `chmod a+x` (unless +x and a+x is already the same, I can never remember)

stavros-k:
Aha, well yes `a+x` will set the execute bits for everyone (user, group, others).
But it will still be owned by root.
Tbh I don't know how it worked so far :P As the `wger` user isn't root. 
I haven't looked deep into to check, so I can't really tell.

But I believe it's best to have it owned by the `user` it will be executed. 
To avoid headaches in the future. With user inconsistencies.

rolandgeider:
ahh, I see what you mean. I'll try to do some tests later today and check that everything still works

stavros-k:
Cool! I noticed this in our CI (TrueCharts Catalog For TrueNAS Scale)
`exec /home/wger/entrypoint.sh: exec format error`
And thought this might be it, but this was running like this for months. 
And I don't see anything changed. Anyway. If I get some time I might try to dive into it myself too

(Not sure if you can see full logs https://github.com/truecharts/charts/actions/runs/3681327880/jobs/6228221329)

stavros-k:
> we had some problems a couple of days ago so that the arm64/amd64 images weren't being built correctly (or only one of them was so that you'd sometimes get the wrong arch), but that is fixed already. As far as I could reproduce it, you'd get the same "format error" when using the _wrong_ image

I see, I'll trigger a CI run to pull latest images

stavros-k:
> > we had some problems a couple of days ago so that the arm64/amd64 images weren't being built correctly (or only one of them was so that you'd sometimes get the wrong arch), but that is fixed already. As far as I could reproduce it, you'd get the same "format error" when using the _wrong_ image
> 
> I see, I'll trigger a CI run to pull latest images

This fixed it, thanks for the heads up

PR #1217, opened 2023-01-03
replace psycopg2-binary with psycopg2

bbkz:
# Proposed Changes

- Although the arm64 image builds, django will crash because the psycopg2-binary is buggy (linked to the wrong library)
- To be able to fix this compiling psycopg2 during image build will do

- [X] Tests for the changes have been added (for bug fixes / features) -> not in django
- [X] Added yourself to AUTHORS.rst -> no changes in django code

### Other questions

  (e.g. database migration)? -> No

### DIFF
--- a/extras/docker/base/Dockerfile
+++ b/extras/docker/base/Dockerfile
@@ -24,6 +24,7 @@ RUN apt-get update \
       python3-pip \
       sqlite3 \
       wget \
+      libpq \
   && rm -rf /var/lib/apt/lists/* \
   && npm install -g yarn sass\
   && locale-gen en_US.UTF-8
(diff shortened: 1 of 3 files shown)


bbkz:
I was focused on my use case using docker images, i am not sure what implication this will have on "normally" installed setups. They may need to install the `libpq5` package with their package manager.

rolandgeider:
Good catch!

Would it make sense to add a requirements_docker.txt for this?

bbkz:
I am not (hin-und-her gerissen) not sure. It would require not to forget to keep them in sync when the requirements changes. Also when someone want's to install it directly on arm (raspberry pi), it would still fail.

I never tried a "non" docker setup but according to the documentation i like the fact that currently only python packages trough pip are required.

bbkz:
I have added a `requirements_docker.txt`.

PR #1405, opened 2023-08-05
PR#1402: Thousand separator change with language selection; Caveat: n…

Brandon-Lu737:
PR#1402: Thousand separator change with language selection; 
Caveat: need to clear the cache when changing the language

# Proposed Changes

- Change the thousand separator according to languages, such as white space for France, comma for England.

- [X] Tests for the changes have been added (for bug fixes / features); *It can be verified by just running it and is very straightforward*

### Other questions

  (e.g. database migration)?
No.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -74,6 +74,7 @@ Developers
 * Bernardo Koen - https://github.com/BernardoKoen
 * Gabriel Liss - https://github.com/gabeliss
 * Alexandra Rhodes - https://github.com/arhodes130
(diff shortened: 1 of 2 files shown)


rolandgeider:
Hi! thanks for the PR, but django already has something like this, we just need to format the data in the template and don't need to maintain a list of countries etc.

https://docs.djangoproject.com/en/4.2/topics/i18n/formatting/#template-filters

Brandon-Lu737:
Ok silly me. That was my first commit. Sure I'll take a look. The Django framework is definitely more effective than the native python.

PR #1408, opened 2023-08-07
add tzdata in docker base image

bbkz:
# Proposed Changes

- add tzdate to docker base image to set time in containers


rolandgeider:
👍🏻

PR #1434, opened 2023-09-23
Update requirements.txt

cicada0007:
# Changes

Added changes in  `requirements.txt` file


rolandgeider:
it seems that the openfoodfacts package did change its api, will look into this this weekend probably

Issue #1438, opened 2023-09-28
Registration with existing email


rolandgeider:
It seems it's somehow possible to register a new user with an existing email. This causes all kinds of problems, e.g. for the email verification

JayanthBontha:
Can I contribute?

PR #1459, opened 2023-10-12
Dupli acc fix

JayanthBontha:
This minor PR would fix the issue #1438.
The problem arises as there is a case-sensitive check for unique email while registration. Emails are case-insensitive.
Changes have been tested in development environment.
# Proposed Changes

- The uniqueness of the email should be checked with a case-insensitive query.

### Other questions

*No extra commands are required.(No migrations)


rolandgeider:
Good catch, thanks!

PR #1514, opened 2023-12-02
EMAIL_USE_TLS or EMAIL_USE_SSL

bbkz:
The default `settings.py` in the docker image has a error and gives the following Message:

```
ValueError: EMAIL_USE_TLS/EMAIL_USE_SSL are mutually exclusive, so only set one of those settings to True.
```

* EMAIL_USE_TLS is for starttls on the submission port
* EMAIL_USE_SSL is for the old dedicated ssl port

# Proposed Changes

- as most mailservers now use the submission port i suggest setting `EMAIL_USE_TLS` as a default.


rolandgeider:
oh good catch

PR #1531, opened 2023-12-17
Fix wrong default setting in docker image

bbkz:
Sorry i made a bad default setting.

Here comes the [fix for docker #74](https://github.com/wger-project/docker/issues/74)


rolandgeider:
I even looked at the django axes' settings page to see what other options there were and didn't see it as well 😄

bbkz:
Yeah.. sorry. As in the helm templates i'm overriding the default, it was working well for me.

PR #1540, opened 2024-01-01
Import Open Food Facts Ingredients without Mongo

strawpants:
# Proposed Changes

Currently,  I understood that downloading the ingredients from the Open Food Facts server requires an import using a running Mongo server. This pull request avoids using Mongo and unpacking the database file. It directly uses the JSONL data dump format as documented [here](https://world.openfoodfacts.org/data) , and loops over the JSON entries extracted from a Gzip stream.

The pull request adds an option to the management command [import-off-products.py](wger/nutrition/management/commands/import-off-products.py).

Furthermore, during the import I've noticed that many entries for `license_authors` were too long for the varchar(600) field so this pull requests also includes a change of this entry to TextField so it can accommodate longer strings. So this pull request does require a database migration.

(I did not do this yet, in light of soliciting your thoughts on this first)

### Other questions
* I did not yet test whether the Mongo import still works as earlier (it is not trivial to test here because wger itself is running inside a docker ecosystem).
  (e.g. database migration)?
Yes, the license_authors  field has changed from varchar(600) to TextField


rolandgeider:
cool idea! How does this work performance/memory wise? (also, my first thought would be to download the dump somewhere very obvious, so that we don't forget to delete it afterwards)

strawpants:
My understanding is that, since the gzipped file is opened as a stream, it doesn't load the entire file into memory. The code just loops over the lines as they are decompressed om the fly.  Also, the downloading itself is done in chunks,  so memory wise I don't expect big issues here. The gzipped archive is kept for now (and not redownloaded if run again). But it could be an option to  automatically remove it after successfully loading the content in the database.  

I didn't try it yet, but ideally, one would do regular (e.g. every week) updates and use the [open food facts delta files](https://world.openfoodfacts.org/data) provided. I suspect they are the same format and since they are much smaller they can be applied incrementally with low effort. In this more advanced setup,  it does make sense to add a new database table so the openfoodfact import events are registered and can be used to decide which files to download.

rolandgeider:
using the delta files would indeed be a huge improvement, at the moment the import is run... very sporadically. I would even suggest that we don't even need a new table, just writing the last date into a text file in the script folder would be enough or something similarly simple.

In any case, I'll try to take a look at your changes and run them on my machine today

wger/nutrition/management/commands/import-off-products.py
```diff
@@ -92,25 +141,25 @@ def handle(self, **options):
...
 
         languages = {l.short_name: l.pk for l in Language.objects.all()}
 
         bulk_update_bucket = []
         counter = Counter()
-
-        for product in db.products.find(
-            {
+        if options['usejsonl']:
+            products=self.products_jsonl(languages=list(languages.keys()),completeness=self.completeness)
```
ebwinters, line 150: is it just my github or is the alignment off here?
image

wger/nutrition/management/commands/import-off-products.py
```diff
@@ -70,15 +71,63 @@ def add_arguments(self, parser):
...
 
+        client = MongoClient('mongodb://off:off-wger@127.0.0.1', port=27017)
+        db = client.admin
+        return db.products.find(filterdict)
+    
+    def products_jsonl(self,languages,completeness):
+        import json
+        import requests
+        from gzip import GzipFile
+        off_url='https://static.openfoodfacts.org/data/openfoodfacts-products.jsonl.gz'
```
ebwinters, line 100: should this be a default in env file? That way if the URL changes, people can update with a hotfix to env file instead of pulling in a new release?

rolandgeider: I was thinking that only "upstream" wger regularly does this and other instances sync from us like with the exercises. Then we don't generate so much traffic for OFF (and don't need to respond as fast to such changes, if they ever occur)

rolandgeider:
Since I was working on a new USDA import, I have merged this branch into mine: #1666
Thanks a lot for this!

Closing here

PR #1561, opened 2024-01-21
Add progress bar to load-online-fixtures

ebwinters:
# Proposed Changes
Adding a progress bar to `load-online-fixtures` command since it is unclear if progress is happening and leads to confusion when running the command as mentioned in [this issue from the docker repo](https://github.com/wger-project/docker/issues/76) 
- Use tqdm package
- Track progress on json download
- Use custom `loaddata` implementation to track progress on `loaddata` command
image

### Other questions

  (e.g. database migration)?
No

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -75,6 +75,7 @@ Developers
 * Gabriel Liss - https://github.com/gabeliss
 * Alexandra Rhodes - https://github.com/arhodes130
 * Jayanth Bontha - https://github.com/JayanthBontha
+* Ethan Winters - https://github.com/ebwinters
 
 Translators
 -----------
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@
 # Application
 bleach[css]~=6.1
 crispy-bootstrap5==2023.10
+tqdm
 
 # this is a fork of django-bootstrap-breadcrumbs
 # we might need to think about migrating away from this completely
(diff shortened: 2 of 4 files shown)


rolandgeider:
Good idea! This could also be useful for other commands like the one that syncs exercises and so. However the download itself is not what takes so long, loading the zip file needs some time as well, but adding some prints after this should at least make clear that things are happening

ebwinters:
@rolandgeider  custom load data handles showing progress on fixture load

rolandgeider:
it does? when I tried it it showed immediately 1/1 and then nothing

ebwinters:
Yeah if you give it a minute to begin that process, it will print progress for both stages (download and dump).

rolandgeider:
This is what I mean, this jumps to 1/1 and then it needs some time to actually import the fixtures: 

Bildschirmfoto 2024-01-23 um 19 58 58

ebwinters:
@rolandgeider  here is what I see
image

So is it not actually tracking fixture import progress correctly?

rolandgeider:
Yes that's what I mean, I'd expect it to show the detailed progress of the import (like processed x/86731 objects)

rolandgeider:
oh wait, doing `loaddata --verbosity=3` does exactly that

ebwinters:
Should I make another pr to default that verbosity?

rolandgeider:
I just commited it 😄

PR #1751, opened 2024-08-13
Allow env specification of Redis connection SSL parameters

taylor-fuller:
# Proposed Changes
I decided to deploy this project in my Kubernetes cluster and it looks like by default there is no way to accept self-signed certificates when using TLS to connect to a Redis instance.

I am currently running a Redis instance on my local network requiring TLS for connections (setting `port 0` and `tls-port 6379` directives in `/etc/redis/redis.conf`) and using a self-signed cert. When attempting to connect to the Redis instance, `django-redis` is throwing the following error: 

`django_redis.exceptions.ConnectionInterrupted: Redis ConnectionError: Error 1 connecting to 192.168.xx.x:6379. [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1007)`

After doing a quick search, I came across [this issue](https://github.com/jazzband/django-redis/issues/554) in `django-redis` that outlines the fix for this. Also, it does appear that configuration is outlined in the `django-redis` repo [README](https://github.com/jazzband/django-redis?tab=readme-ov-file#ssltls-and-self-signed-certificates) as well.

While I was adding this I figured it would be good to also include the ability to specify other possible common SSL configuration parameters as outlined [here](https://github.com/redis/redis-py/blob/master/redis/connection.py#L743-L760) in the `redis-py` project.

Assuming these changes are acceptable and once they are approved/merged I will open an accompanying PR in https://github.com/wger-project/docker to document these changes as well.

- ~~[ ] Tests for the changes have been added (for bug fixes / features)~~ N/A given this is an env configuration

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -79,6 +79,7 @@ Developers
 * Ethan Winters - https://github.com/ebwinters
 * Dieter Plaetinck - https://github.com/Dieterbe
 * Jonathan La Field - https://github.com/JLaField
+* Taylor Fuller - https://github.com/taylor-fuller
 
 
 Translators
--- a/extras/docker/development/settings.py
+++ b/extras/docker/development/settings.py
@@ -111,6 +111,12 @@ if os.environ.get("DJANGO_CACHE_BACKEND"):
             'TIMEOUT': env.int("DJANGO_CACHE_TIMEOUT"),
             'OPTIONS': {
                 'CLIENT_CLASS': env.str("DJANGO_CACHE_CLIENT_CLASS"),
+                'CONNECTION_POOL_KWARGS': {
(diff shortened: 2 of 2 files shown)


rolandgeider:
(sorry for the late reply). I'm guessing if the values are not set they are ignored right? In that case it, could you also open a PR on the docker repo adding these options as a comment to the env file so it's clear they can be set?

taylor-fuller:
> (sorry for the late reply). I'm guessing if the values are not set they are ignored right? In that case it, could you also open a PR on the docker repo adding these options as a comment to the env file so it's clear they can be set?

@rolandgeider 

All good! 

I went ahead and made a slight adjustment so `CONNECTION_POOL_KWARGS` is only added to the `CACHES` config if at least one of the related env parameters is present.

Also, please see [accompanying PR for the docker repo](https://github.com/wger-project/docker/pull/98)

Issue #1797, opened 2024-10-18
Authentication: Add OIDC Integration


redxtech:
## Use case

I would like to see OIDC support for authentication, to make it easier to integrate with existing setups that use auth providers such as authentik.

## Proposal

Django appears to have support for OIDC build in (https://django-oidc-provider.readthedocs.io/en/master/), so adding it here should definitely be easier than if you were to start from scratch. If that somehow doesn't apply here, there should probably be another way to implement this as well. Let me know if I'm wrong!

kmoy1:
Would like to work on this

kmoy1:
Is part of this request to add an OIDC authentication backend and authenticate against that instead of the default `django.contrib.auth.backends.ModelBackend` (checking against the User model)? If that's the case, I am not sure `django-oidc-provider` provides one- may need to use the [mozilla-django-oidc](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html?highlight=authentication_backend) library instead which does.

rolandgeider:
We definitely need a user object for the rest of the application to work. I'm not sure how the oidc provider works, but I assumed it would do that automatically

kmoy1:
Yes, I was just thinking that if we set up an OIDC IDP like Authentik, we would manage user accounts there instead of locally in database.sqlite3 like we do currently. Unless I'm mistaken?

kmoy1:
I've added a basic example of support for OIDC authentication using the `mozilla-django-oidc` library. However, this CR does require the manual setup of an OIDC provider (I used Authentik here). I didn't see another way around this.

InputObject2:
If we're setting up oidc auth, the oidc provider part isn't wger's responsibility. When you say "require the manual setup" did you mean we can no longer use "local" auth?

Also needing a couple environment variables is pretty common. These are well-known variables that most providers use, so it wouldn't be limited to working with Authentik.

rolandgeider:
This only allows to setup one OIDC provider right? Is that enough or would we potentially want more?

kmoy1:
Yes, this was meant specifically for the Authentik. But the mozilla-django-oidc library can definitely extend to other OIDC providers too. Again, the issue is that (as far as I know) the provider needs to be configured manually and separately from this app.

LeVraiRoiDHyrule:
Hi, is there news on this ? Auth headers works right now but true OIDC would be even better.

Keridos:
[...]

PR #1805, opened 2024-11-01
Add basic OIDC integration with locally setup Authentik

kmoy1:
# NOTE: Since this requires local manual setup of an OIDC IDP, this should not be merged into master but into its own branch (e.g. 'OIDC').

Addresses #1797 

# Proposed Changes

- Added a very basic form of OIDC Authentication support, by adding a "Login with Authentik" button to the Login page. It will authenticate with another (locally) spun-up Authentik IDP server, instead of requiring username/password combo. It will use the email of the user instead. 

- [] Tests for the changes have been added (for bug fixes / features) - not sure how to test this, because this code review requires manual setup of an OAuth2 provider. 

### Other questions

  (e.g. database migration)?

Yes. Users need to spin up an Authentik application of their own and then set these environment variables in the Django app (or hardcode them in settings_global.py: not recommended), values of which are provided by Authentik. 

OIDC_RP_CLIENT_ID
OIDC_RP_CLIENT_SECRET
OIDC_OP_AUTHORIZATION_ENDPOINT
OIDC_OP_TOKEN_ENDPOINT
OIDC_OP_USER_ENDPOINT
OIDC_OP_JWKS_ENDPOINT

They also need to set 
OIDC_RP_SIGN_ALGO=RS256

### DIFF
--- a/wger/core/forms.py
+++ b/wger/core/forms.py
@@ -39,6 +39,7 @@ from django.utils.translation import gettext as _
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import (
     HTML,
+    Button,
     ButtonHolder,
     Column,
     Fieldset,
@@ -51,6 +52,10 @@ from django_recaptcha.widgets import ReCaptchaV3
 
 # wger
 from wger.core.models import UserProfile
+
+#OIDC
+from django.urls import reverse
+from django.shortcuts import redirect
 
 
 class UserLoginForm(AuthenticationForm):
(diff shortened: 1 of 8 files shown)


Keridos:
This looks like a draft for me. Under the assumption that the implementation is working:
The OIDC login should be only activated when the OIDC Variables are set.
Additional Choice: When it is activated you may disable regular account login with username and password.
Additionally it would be very nice to make the name of the OIDC Provicer (on the button) also configurable via env variable.

Sadly I lack the skills for web development and never worked with Django so far or I'd implement the changes myself.

Issue #1803, opened 2024-10-25
Setting preferred weight unit to "Imperial (pound)" still defaults to kg for exercise units when creating workouts


slipperybeluga:
## Steps to Reproduce

1. Set preferred weight unit to Imperial (pound) in My preferences
2. Go to Workouts, select one. Click the "+" button to add an exercise to a workout day
3. Select a lift (eg Bench Press). Set reps, etc. Now have to manually update unit from kg to lb for each rep. 

**Expected results:** Default unit would be lb

**Actual results:** Default unit is kg

kmoy1:
looking into this

rolandgeider:
This is done in via the linked PR as well as in the coming refactoring

PR #1807, opened 2024-11-02
Dynamically update WeightUnit from user preferences when creating workout

kmoy1:
# Proposed Changes

Address issue #1803 by dynamically updating WeightUnit from user preferences when creating workout formset. 

In the `get_import` view, we query the UserProfile model for the preferred weight unit (set in preferences). Then, we have to pass in a custom form class (SettingFormOverride) to override the initial value of weight_unit to our preference.

### Other questions

  (e.g. database migration)?

No


rolandgeider:
thanks for the PR, and props to you, that is one of the hackiest, most ugly parts of the whole application! That will change once I finally finish the flexible routines, but this is an improvement till then

PR #1833, opened 2024-12-05
make __main__ use pyinvoke `Program` as entrypoint

eyJhb:
This is conforms more to the recommended way of making a pyinvoke project.
Inspiration was generally taken from fabric, e.g. https://github.com/fabric/fabric/blob/main/tasks.py, which gives a good example of this.

This change is required for packaging for e.g. nix, as the other method called the `invoke` binary, which changes how paths works. Now it's all contained INSIDE this project, and doesn't call external binaries. 

Should be good to go. Hopefully.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -81,6 +81,7 @@ Developers
 * Jonathan La Field - https://github.com/JLaField
 * Kevin Moy - https://github.com/kmoy1
 * Taylor Fuller - https://github.com/taylor-fuller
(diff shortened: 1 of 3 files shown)


rolandgeider:
merged, thanks

eyJhb:
Sorry about the auto format, I hadn't noticed that. I'll be more careful in the future! :)

rolandgeider:
I had tne branch already checked out, so I just fixed it. In any case the code would be auto linted once it landed on master

PR #1859, opened 2025-01-02
adds support for auth proxy header

### LINKED ISSUE #1847: Support for Proxy Auth via. HTTP header
## Use case
This would allow to put e.g. Authelia in front of Wger, and then use the header that Authelia passes to wger to create/signin the user.
This could function as a simple way, to  support the uses cases such as #1797 (not 100%, but it is quite simple to make).

See https://github.com/miniflux/v2/issues/534 and https://grafana.com/docs/grafana/latest/auth/auth-proxy/ .

## Proposal
- Add config option `PROXY_AUTH_HEADER=Remote-User`
- Add config option `PROXY_AUTH_CREATE_USER=True`
- `ALLOW_REGISTRATION=False` should still allow creating users using `PROXY_AUTH_`
- Mobile app should allow using API key instead of username/password for signin
  - This means user would first authenticate themselves using the website (go through e.g. Authelia -> Wger), copy API key, and then use that from the app.

I might take a jab at this if I have the time.

eyJhb:
Useful if you have authentication in front of wger, and want to use that instead of wgers authentication/signup methods.

# Proposed Changes

Added a new setting, which can be used to setup proxy auth header, e.g. using Authelia to authenticate + create users on the instance.

Fixes #1847 , and potential VERY SIMPLE solution for #1805 ( #1797 )

### Other questions

  (e.g. database migration)? No

# Remaining work

- [ ] Add to documentation.
- [ ] Flutter needs to  support setting API key, instead of username/password.


rolandgeider:
BTW I haven't forgotte about this PR, but since it would add a new auth method I'd like to make sure we do it right. I've added this to the 2.4 backlog (so after the flexible routines get merged)

PR #1875, opened 2025-01-11
Add language filter to sync-ingredients management command.

navyjosh:
# Proposed Changes

- Allow user to specify a language when syncing ingredients
- example: `python manage.py sync-ingredients --languages=en,fr,es`

  - `wger.nutrition.tests.test_xync.TestSyncMethods` has been slightly modified to ensure uri was called with proper filter.

### Other questions

  (e.g. database migration)? No

### DIFF
--- a/wger/nutrition/sync.py
+++ b/wger/nutrition/sync.py
@@ -15,7 +15,7 @@
 # Standard Library
 import logging
 import os
-from typing import Optional
+from typing import List, Optional
 
 # Django
 from django.conf import settings
@@ -29,6 +29,7 @@ from openfoodfacts.images import (
 )
 
 # wger
+from wger.core.models.language import Language
 from wger.nutrition.api.endpoints import (
     IMAGE_ENDPOINT,
     INGREDIENTS_ENDPOINT,
@@ -48,6 +49,7 @@ from wger.utils.requests import (
     get_paginated,
     wger_headers,
 )
+from wger.utils.language import load_language
 from wger.utils.url import make_uri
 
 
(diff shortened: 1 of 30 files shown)


wger/nutrition/tests/test_sync.py
```diff
@@ -101,9 +101,9 @@ def test_ingredient_sync(self, mock_request):
         self.assertEqual(ingredient.code, '1234567890987654321')
 
         # Act
-        sync_ingredients(lambda x: x)
+        sync_ingredients(lambda x: x, language_codes='en')
         mock_request.assert_called_with(
-            'https://wger.de/api/v2/ingredient/?limit=999',
+            'https://wger.de/api/v2/ingredient/?limit=999&language=2',
```
rolandgeider, line 106: I'd add a new simple test here, something like `def test_ingredient_sync_languages` that only checks that the correct URL is called

wger/nutrition/sync.py
```diff
@@ -216,40 +221,73 @@ def download_ingredient_images(
...
     style_fn=lambda x: x,
 ):
     """Synchronize the ingredients from the remote server"""
+
+    def _sync_ingredients(language_id: Optional[int] = None):
+        if language_id is not None:
+            url = make_uri(
+                INGREDIENTS_ENDPOINT,
+                server_url=remote_url,
+                query={'limit': API_MAX_ITEMS, 'language': language_id},
```
rolandgeider, line 234: I'd just allow passing all languages at once, seems easier that way. For this we just need to add 'in' to `IngredientFilterSet`, like in 'id'. Then we can do `query={'limit': API_MAX_ITEMS, 'language__in': ','.join([languages])}` (this also works if we only pass one)

Issue #1814, opened 2024-11-10
Allow more than one weight entry per day


rolandgeider:
At the moment, it's only possible to add one weight or measurement entry per day. While it doesn't make much sense to log this several times per day, it also doesn't hurt and it would make some of the logic more easy

Dieterbe:
A trainer on our discord brought up a good use case: water cuts for making weight categories (eg in powerlifting comp)

scrapcode:
I am taking a stab at this but I think clarification needs to be made on any desired restrictions, or none at all.

Currently there are many tests that check that seed data helper functions filter out weight entries for existing dates, date addition functions update existing entries when made within N days, etc.

Are there any restrictions desired here, for example, only one weight entry per day? Or remove such restrictions altogether?

rolandgeider:
just remove them alltogehter :)

PR #1895, opened 2025-02-23
Allow multiple weight entries per day

scrapcode:
# Proposed Changes

- Refer to Issue #1814 
- Changed the `WeightEntry` date attribute from a `DateField` to a `DateTimeField`
- Removed the `unique_together` of user and date.
- Removed the restriction of only being allowed to add a new weight entry every so often. You can now add multiple weight entries per day.
   - Removed or altered tests that tested this functionality
- Changed the method of getting and constructing dates from naive to aware, and to include times.
- Added DATETIME_FORMATS constants.
- Modified seeded testing data to reflect aware datetime formats.

### TODO: Frontend Implementation
- This will still need implementation on the frontend. Adding multiple weight entries per day works but are currently all added with the time as `23:00:00`. How this is displayed on the graph may want to be taken into consideration as well.
   - flutter#641
   - react#990

   - Multiple tests and testing methods had to be modified or removed to allow for the new functionality.

### Other questions

  (e.g. database migration)?
 - `python manage.py migrate`


rolandgeider:
you're on fire!

Also, there were more changes needed that I had thought. Will you take a shot at the frontend as well or should I do it?

scrapcode:
>  Will you take a shot at the frontend as well or should I do it?

I'll give it a whirl!

rolandgeider:
we can finally merge this 🥳 (and sorry for the wait)

PS: the last test is failing because it can't upload the coverage report, so all is fine

Issue #1956, opened 2025-04-14
Handle special characters in OFF product import


rolandgeider:
some ingredients from OFF have names like e.g.  "&quot;Phnompenh&quot; Rice Stick".

TODO:
- find some other problematic characters (just the most common ones)
- fix the OFF import

wbonicki:
Hello, can I take care of that?

rolandgeider:
Hi! Sure, go ahead!

The script that imports the OFF db is `wger/nutrition/management/commands/import-off-products.py`. You should download the JSONL dump and pass the path with `--folder`, otherwise the script will download and delete it on every run (I'm not sure right now if you also have to extract it): https://world.openfoodfacts.org/data (the JSONL data export)

And heads up, the dump is several GBs of compressed data, once extracted this easily reaches like 35 GBs

wbonicki:
Hi, I did some analysis: I read all the entries from OFF, created a set of all characters found in the names (the size of the set for the processable entries is 2,831), and then categorized the characters as follows:

Cc – Control characters
Cf – Format characters (e.g., zero-width space, RTL marks)
Zs – Space separators (e.g., non-breaking space, em-space)
So – Symbols and emoji
Quotation/parenthesis – Various quotation marks and parentheses

The results are as follows:

{'Cc': ['\x8a','\x83','\x8f','\x8d','\x82','\x9a','\x89','\x99','\x87','\x9c','\x80','\x96','\t','\x92','\x8c','\x90','\x9f','\x84','\x97',x93','\x98','\x85'],
 'Cf': ['\u200f', '\xad', '\u200d', '\u202b', '\u200b', '\u2060'],
 'Quotation': ['“', '〈',')', '‘', '』', '（', '[', '》', '{', '„', '『', '《','‚', ']', '}', '］', '»', '〉', '”', '）', '［', '«', '(', '’'],
[...]

PR #1970, opened 2025-05-12
Handle special characters in OFF product import #1956

wbonicki:
# Proposed Changes

- added handling for special characters (removed) and HTML entities (converted to human-readable form) in product names during the OFF import

## Related Issue(s)

https://github.com/wger-project/wger/issues/1956

### Other questions

* the `import-off-products` command must be executed to update and clean existing product names in the database.


wbonicki:
Is it my responsibility to fix this error (Lint Code Base)?

rolandgeider:
no, no, I'm not sure why that error happens, you can just ignore it

wger/utils/constants.py
```diff
@@ -77,3 +77,77 @@
...
+        '\x97',
+        '\x93',
+        '\x98',
+        '\x85',
+    ],
+    'format_characters': [],
+    'space_separators': [],
+    'quotation_or_parenthesis': [],
+    'symbols_and_emoji': [],
+}
```
rolandgeider, line 110: if we're only using the control chars, we could just replace the dict with a simple list?

wbonicki: done

rolandgeider:
merged, thank you!!

PR #1975, opened 2025-05-22
Improve logging for ingredient sync operations

### LINKED ISSUE #1896: Reduce logging in sync-ingredients
At the moment when running `manage.py sync-ingredients` every created or update ingredient is logged to the console. While this is very informative, when synchronizing the whole database this results in 2.5 million log messages...

So:
* only output them when the verbose option is set, this option available to all management commands
* check if we can just use tqdm to show the progress

rolandgeider:
# Proposed Changes

Don't log every single updated or created ingredient anymore. We now use tqdm to show a nice progress bar when calling this via the command line, and a more general "page x of y" when using the async method

## Related Issue(s)

Closes #1896

### DIFF
--- a/wger/nutrition/management/commands/sync-ingredients.py
+++ b/wger/nutrition/management/commands/sync-ingredients.py
@@ -50,11 +50,16 @@ class Command(WgerCommand):
 
         check_min_server_version(remote_url)
         try:
-            self.languages = languages
-            if self.languages is not None:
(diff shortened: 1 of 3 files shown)


rolandgeider:
Bildschirmfoto 2025-05-22 um 11 44 53
Bildschirmfoto 2025-05-22 um 11 48 16

PR #2030, opened 2025-08-29
Prevent height spinner going below 0 in user preferences

Carner-Lin:
# Proposed Changes

- Set min="0" (and type="number", step="1") on UserPreferencesForm height widget so the down arrow can not go below 0. 
- There is also a default prompt when the keyboard enters a negative number.

### Other questions

- None


rolandgeider:
thanks for the PR! (and sorry for the wait)

Issue #2047, opened 2025-09-21
Malformed URL for self-hosted email verification


kmanwar89:
### Priority/Impact

Medium (affects some functionality)

### Description

After cloning the repository and editing the `prod.env` environment variables to enable the SMTP email service through gMail, I encounter the following in my docker compose logs:

`web-1            | 2025-09-21T12:37:47.868564779Z level=WARNING ts=2025-09-21 08:37:47,868 module=confirm path=/home/wger/.local/lib/python3.12/site-packages/django_email_verification/confirm.py line=82 message=WARNING: the URL seems to be malformed - http://localhost/email/email/eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImthZGFyYW53YXIwN0BnbWFpbC5jb20iLCJleHAiOjE3NTg0NjE4NjcuODY2NDY1LCJraW5kIjoiTUFJTCJ9.ShrZcZzvaALARxeHub4dH8VSE4w3mSJLlH9P9O2E5fQ`

As a result, clicking the link in the verification email FAILS. The URL that shows up in the web browser is:

`http://localhost/email/email/eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImthZGFyYW53YXIwN0BnbWFpbC5jb20iLCJleHAiOjE3NTg0NjE4NjcuODY2NDY1LCJraW5kIjoiTUFJTCJ9.ShrZcZzvaALARxeHub4dH8VSE4w3mSJLlH9P9O2E5fQ`

Obviously, this won't work with the redirect to localhost.  I found that a workaround is to replace "localhost" in that URL with the actual URL of the service (subdomain.domain.com)

Is it possible to add an environment variable that can be set to the subdomain and substituted automatically?

### Server version

Server version: 2.4.0a1; installed via docker compoose cloned from repo on GitHub

KatharinaJacoby:
Happy Hacktoberfest! I am on it

rolandgeider:
Hey @KatharinaJacoby thanks for the PR, but SITE_URL already does this, no need to add a new one :)

[...]

PR #2060, opened 2025-10-01
Email verification: use PUBLIC_APP_URL for absolute links

KatharinaJacoby:
Fixes wger-project/wger#2047

When self-hosting, verification emails previously contained links like
`http://localhost/...`, which break when running behind a reverse proxy.

This PR introduces a new environment variable `PUBLIC_APP_URL` that, if set, is
used for building absolute verification links instead of falling back to
`SITE_URL`.

### Changes
- settings.tpl: support for PUBLIC_APP_URL (normalized, sets EMAIL_PAGE_DOMAIN)
- extras/docker/production/settings.py: same support for production deployments
- extras/docker/production/README.md: documented the new variable under a Configuration section

### Notes
- If PUBLIC_APP_URL is not set, behavior remains unchanged (falls back to SITE_URL).
- No database migrations.
- Backward compatible with existing installs.

### DIFF
--- a/extras/docker/production/settings.py
+++ b/extras/docker/production/settings.py
@@ -2,6 +2,7 @@
 
 # Third Party
 import environ
+import os
 
 # wger
 from wger.settings_global import *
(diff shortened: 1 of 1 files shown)


rolandgeider:
Closing here, see comment in original issue

PR #2066, opened 2025-10-02
Fix "Get Started" Button and Add Registration Sidebar

### LINKED ISSUE #2014: "Get Started" button on live site behavior
### Use case

I was checking out the live site and I noticed that when you click "Get Started" , it redirects the user to the login page. I think it would make more sense if this "Get Started" button would lead to the registration page.

### Proposal

I think that this change is needed because it is more consistent with what a user would expect when clicking the "Get Started" button on the landing page. The problem that it solves is just that it would lead to less confusion for a new user and also, it would help the UI of the webpage be more consistent with common usability heuristics (or expected behaviors) such as those indicated here: https://www.nngroup.com/articles/ten-usability-heuristics/

I'd like to take care of this issue, if possible.

### Additional Context

There are a few possible solutions that I thought of. We could either make the button simply link to the registration page, or the login page can have a link to register an account.

Landing page: Note the "Get Started" button.
Image

Proposed Change for login page: note the added "New user?" 
Image

JANVI-CHATURVEDI:
## Proposed Changes

* Fixed the "Get Started" button on the live site to redirect to the registration page instead of login. (#2014)
* Added a **registration-specific sidebar** template (`registration_sidebar.html`) that shows the "Already have an account?" link for anonymous users on the registration page.
* Updated `form_content.html` to include the new sidebar only when specified.
* Corrected the "Get Started" link on `features.html` to point to the appropriate page depending on user authentication.

## Related Issue(s)

* Closes #2014

## PR Checklist

* [x] My changes include only the intended updates.
* [x] No unnecessary files, migrations, or large diffs are included.
* [x] Code follows the repository's formatting and style guidelines.

## Additional Notes

* No database migrations or extra commands are required.
* The `{% if allow_registration %}` logic on the login page remains unchanged as per maintainer guidance.
* This PR uses a clean branch including only the relevant changes (`fix-getstarted-clean-final`).

### DIFF
--- a/wger/core/templates/form_content.html
+++ b/wger/core/templates/form_content.html
@@ -23,7 +23,4 @@
 {% endblock %}
 
 
-<!--
-        Sidebar
--->
 {% block sidebar %}{% if sidebar %}{% include sidebar %}{% endif %}{% endblock %}
--- a/wger/core/templates/user/registration_sidebar.html
+++ b/wger/core/templates/user/registration_sidebar.html
@@ -1,0 +1,17 @@
+
+{% load i18n static wger_extras crispy_forms_tags %}
+
+
+
+{% block sidebar %}
+{% if user.is_anonymous %}
+    <h4>{% trans "Already have an account?" %}</h4>
+    <p>
+        <a href="{% url 'core:user:login' %}">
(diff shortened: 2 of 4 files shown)


rolandgeider:
🚀

Issue #2032, opened 2025-09-02
Possibility to add custom exercise names.


LepaBrenaa:
### Use case

I'm trying to add "Lat Pulldown" as an exercise, but I'm forced to choose the exercise names that are being suggested, and none of them are what I want.
I just want to be able to name the exercise however I want instead of being forced to choose the name out of some database.

### Proposal

Let users manually type the name of the exercise instead of forcing them to choose the name from a database.

Katzi23:
Hello, I started working on this here: https://github.com/Katzi23/wger/tree/CUSTOM_EXERCISES

rolandgeider:
@Katzi23 no wait! I'm not sure we're going this route of allowing custom exercises

Katzi23:
Oh that's ok! You can just ignore it then

elias170105:
What's wrong with custom exercises?
In the end, you can offer it as an option. The advantage would be that the app would be helpful to even more people and everyone could customize the app.

PR #2069, opened 2025-10-03
Custom exercises

Katzi23:
…es, migrations)

# Proposed Changes

- Backend setup for users to add their own custom exercises

## Related Issue(s)

Related to wger-project/wger#2032

### Other questions

Yes — this PR includes database migrations; after pulling, users should run “python manage.py migrate.”

This is my first contribution so sorry if it's messed up!


rolandgeider:
forgot to comment here. Sorry you already did all this work!!! As you can see, you had to touch **a lot** of files and add special cases everywhere to make this work (and there's still the mobile app and the react code we'd also need to touch in one way or the other). While I understand that people have their favourite exercise names, I don't think this warrants it. We already have some aliases for them, perhaps a solution in the future would be to mark one of those as the favourite and don't see the "official" name, but I'm not even sure about that. In any case, thanks again for the PR, even if I have to close it now!

PR #2087, opened 2025-10-17
Fix login bug causing incorrect credential handling

abbadinikhilreddy55:
### What I changed
Fixed an issue where the login flow did not properly handle incorrect user credentials, resulting in unexpected errors or no error messages shown.

### Why I changed it
Users were unable to see appropriate error feedback when entering wrong login details, which caused confusion and prevented proper authentication.

### How I tested it
- Manually tested login with valid and invalid credentials to verify correct behavior.
- Ran existing automated tests; all tests passed successfully.
- Verified that error messages display correctly in the UI when login fails.

### DIFF
--- a/edit file
+++ b/edit file
@@ -1,0 +1,363 @@
+
+# -*- coding: utf-8 -*-
+
+# This file is part of wger Workout Manager.
+#
+# wger Workout Manager is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


rolandgeider:
hi! what do you mean that there is no feedback? If the password doesn't match there's an info box

Bildschirmfoto 2025-10-17 um 15 45 18


Also, did you commit the wrong file?

rolandgeider:
@abbadinikhilreddy55 sorry there seems to be some misunderstanding. What exactly did you change? The PR only adds a new file "edit file" that seems to be a copy of the urls

Issue #2134, opened 2025-12-02
Include a timer / clock for those who do intermittent fasting


dovy6:
### Use case

I've been using the app called Zero for the past ~6 years to track my fasting times. I'd love to be in control of my own data again. 
I'm looking for a button that says "start fasting" when you aren't, "end fast" when you are, a timer that counts how long you have been fasting for (or counts how much longer until you can eat, depending on preferences), as well as a history of prior fasts.

### Proposal

Thanks so much for considering! Let me know if you need any further info

### Additional Context

<img width="1008" height="2019" alt="Image" src="https://github.com/user-attachments/assets/e191ef09-45e8-4a4c-8022-0bad6ec16740" />

gracechen636:
Hi! I am interested in looking into this

rolandgeider:
Hey @gracechen636 , sure go ahead!

@dovy6 how does this work in zero? I'm guessing there it's a global switch for fasting / not fasting right? In our case it would probably be much easier to put this info in the nutritional plan object. Also, is there only one period for fasting or can there be more?

dovy6:
@rolandgeider are you asking how it works on the backend, or from a user experience? I'm not a developer, and I imagine that the Zero app isn't open source regardless, but I could tell you how it works on my end.

When I'm finished eating for the day, I press the large "start fasting" button, which starts a timer (in my case counting up from zero, but it can be configured to count down from your assigned fasting period as well). When I'm ready to eat again, I press the same button (now labeled "complete fast") which turns the timer into a "time since last fast" counter.

About periods, do you mean fasting periods? It is configurable. Mine is currently set to 16:8, though I generally average more like 20:4 nowadays.

Please let me know if and how I can assist!

rolandgeider:
I meant from the user's point of view 😄 

Yes, fasting periods. I was asking in case it was usual to have more than one.

dovy6:
Some people do have varying fasting periods, for example 16:8 5 days a week and 12:12 on weekends. Personally, I leave mine set to 16:8 even though as I mentioned I usually go 20:4 or even 22:2 most days.

PR #2138, opened 2025-12-09
Add fasting feature: FastingWindow model, tests, and views

gracechen636:
Proposed Changes

Add FastingWindow model to represent user fasting periods, including start/end times, optional goal duration, etc.

Provide convenience properties for active status and duration calculations.

Add unit tests covering model creation, behavior, and ordering.

Related Issue(s)
#2134 

Do users need to run some commands in their local instances due to this PR

Maybe.
This PR introduces a new model, so users may need to run database migrations:

python manage.py makemigrations
python manage.py migrate


rolandgeider:
NB: I'm planning to finish implementing the UI for the trophies, doing a release, and then merging this for the next one

PR #2160, opened 2026-01-10
Feature: set UI language

### LINKED ISSUE #2121: Allow explicitly setting the UI language
### Use case

From discussion https://github.com/wger-project/wger/discussions/2085:

Users that have one system language but want to use the app in another need to always change it. It seems it's possible to use a cookie that will be used instead of browser headers: https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#miscellaneous

### Proposal

Add a simple dropdown that saves the selected language to the django cookie

### Additional Context

_No response_

franchsli:
# Proposed Changes

**Summary of changes**
- Explicitly set UI language using the languages dropdown in the footer. Users don't need to set the language every time they log in now.
- Let users use their browser's language if they want to by clicking in the "Browser language" link in the dropdown.

**Implementation details**
- The languages in the dropdown set a django cookie to store the language.
- The "Browser language" link calls a view that deletes the cookie and redirects the user to the current page without language prefix so Django uses the browser's language.

**Extras**
- `remove_language_code` helper function was fixed to handle region-specific locales like "es-ar" or "zh-hans" which are present in wger's supported languages.

## Related Issue(s)

Closes #2121 

### Other questions

A// No, they don't.

### DIFF
--- a/wger/core/templatetags/wger_extras.py
+++ b/wger/core/templatetags/wger_extras.py
@@ -74,7 +74,7 @@ def language_select(context, language):
     Renders a link to change the current language.
     """
 
-    return {**get_language_data(language), 'i18n_path': context['i18n_path'][language[0]]}
+    return get_language_data(language)
 
 
 @register.filter


rolandgeider:
thanks for the PR!

I want to finish the work on the trophies and then do a release. I'll take a look at this afterwards

rolandgeider:
BTW, there's also a language switcher in `wger/software/templates/features.html`, that would need to be edited as well, but then we can merge this

We also don't need `i18n_path`, which was previously used in `language_select.html`, so we can also remove it from `wger/utils/context_processor.py`

franchsli:
> BTW, there's also a language switcher in wger/software/templates/features.html, that would need to be edited as well, but then we can merge this

Oh, I totally missed it, I'll apply the changes to it.

> We also don't need i18n_path, which was previously used in language_select.html, so we can also remove it from wger/utils/context_processor.py

Ok, I'll remove it then

rolandgeider:
Merged, thanks!

PR #2181, opened 2026-01-23
Fix: Auto-calculate SlotEntry.order when not provided via API

ndimoro:
## Summary

- Fix 500 Internal Server Error when creating SlotEntry via `/api/v2/slot-entry/` without specifying `order`
- The `order` field was defined with `blank=True` but no default value, causing `IntegrityError: null value in column "order" violates not-null constraint`

## Changes

1. Added `null=True` to the `order` field to allow Python `None` value before save
2. Added auto-calculation logic in `save()` method: `max(existing_orders) + 1`

This matches the behavior of similar models (`Day`, `Slot`) which have `default=1` for their order fields.

## Error before fix

```
psycopg.errors.NotNullViolation: null value in column "order" of relation "manager_slotentry" violates not-null constraint
django.db.utils.IntegrityError: null value in column "order" of relation "manager_slotentry" violates not-null constraint
```

## Test plan

- [x] Tested creating SlotEntry without providing `order` - auto-calculated correctly
- [ ] Run existing test suite

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/wger/manager/models/slot_entry.py
+++ b/wger/manager/models/slot_entry.py
@@ -140,7 +140,6 @@ class SlotEntry(models.Model):
 
     order = models.PositiveIntegerField(
         blank=True,
-        null=True,
         db_index=True,
     )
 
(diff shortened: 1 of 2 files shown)


rolandgeider:
thanks for the PR! I've removed the `null=True` from the model so in the db there's always a value (which we now compute if it's not available)

PR #2192, opened 2026-01-31
add brand and common name on ingredientinfo serializer

### LINKED ISSUE #2191: Show brand on ingredient info api
### Use case

Hi, can we show brand field in /api/v2/ingredientinfo/, so later on in mobile apps can show it too

### Proposal

Sometimes we have same ingredient name but with different nutrition that came from different brand, is will be useful if we can show that brand name when user want log their food, so they will know exactly which ingredient that they want to add

### Additional Context

_No response_

cryptbrn:
# Proposed Changes

- Add brand and common_name field on ingredientinfo serializer

## Related Issue(s)
- closes #2191

### DIFF
--- a/wger/nutrition/api/serializers.py
+++ b/wger/nutrition/api/serializers.py
@@ -166,6 +166,8 @@ class IngredientInfoSerializer(serializers.ModelSerializer):
             'source_url',
             'code',
             'name',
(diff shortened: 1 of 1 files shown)


rolandgeider:
merged, thanks

PR #2222, opened 2026-02-26
Add is ai generated flag to images

alexisbirling-boop:
# Proposed Changes

- Add `is_ai_generated` boolean field to `ExerciseImage` model to track whether an image was generated by AI
- Expose the new field through the `ExerciseImageSerializer` in the REST API
- Include field in `HistoricalExerciseImage` model for audit trail consistency
- Set default value to `False` for all existing and new images

## Related Issue(s)

This was discussed with rge on Discord directly. 

## Screenshot of the API

ai-flag

### DIFF
--- a/wger/exercises/migrations/0035_add_is_ai_generated.py
+++ b/wger/exercises/migrations/0035_add_is_ai_generated.py
@@ -1,10 +1,9 @@
-﻿# Generated manually
+# Generated manually
 
 from django.db import migrations, models
 
 
 class Migration(migrations.Migration):
-
     dependencies = [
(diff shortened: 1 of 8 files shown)


rolandgeider:
thanks for the PR.

BTW, the tests for python 3.13 are failing because the service we use to upload the coverage report is currently down

rolandgeider:
one thing that is still missing, we need to read this new field when performing exercise syncs. Just add it to `wger.exercises.models.image.ExerciseImage.from_json` (best if we do a `json_data.get()` so that this works even if people have older server versions etc)

rolandgeider:
merged thanks!

PR #2259, opened 2026-03-21
feat: Allow AUTH_PROXY_TRUSTED_IPS to accept subnetfeat: Allow AUTH_PROXY_TRUSTED_IPS to accept subnet via ipaddress module (#1989) (#1989)

### LINKED ISSUE #1989: Allow `AUTH_PROXY_TRUSTED_IPS` to accept subnet instead of just single IP
### Use case

Allow `AUTH_PROXY_TRUSTED_IPS` to accept subnet instead of just single IP.

### Proposal

> - It would be nice if `AUTH_PROXY_TRUSTED_IPS` accepted a subnet. Using podman, all of the wger containers in the pod are part of their own network. I trust all the containers in this subnet. The container corresponding to caddy is not guaranteed to have the same IP each time the pod is started (and I don't want to specify a static IP because the less hard-coded values the more maintainable the configuration is).
> 
> Otherwise, thank you for implementing this feature and thank you for the wger project!!! I am using it in my home cloud. 

 _Originally posted by @tricepsbrachii in [#1847](https://github.com/wger-project/wger/issues/1847#issuecomment-2932504660)_

> Thanks to the helpful information on this page I was also able to get it working. 
> 
> I also second that it would be nice to set a subnet for the trusted IP. My entire setup will break if the docker ip of the reverse proxy changes. For now I will manually set the up for the reverse proxy in the compose file, but that's not a very good solution IMHO. 

 _Originally posted by @Gecko-with-a-hat in [#1847](https://github.com/wger-project/wger/issues/1847#issuecomment-2966712083)_

As others and myself have pointed out before, it would be great if we could set the `AUTH_PROXY_TRUSTED_IPS` env variable to a subnet instead of just a single ip address. Currently it's not the best solution when you rely on docker networks for your reverse proxy. Would be great to be able to set it to ```172.20.0.20/24``` or ```10.0.0.0/8``` with CIDR notation. This would IMO greatly improve compatibility and usability.

### Additional Context

> I also second that it would be nice to set a subnet for the trusted IP.
> 
> this sounds like a nice addition and should be easy enough to check with the `ipaddress` package. @Gecko-with-a-hat can you open a new issue? 

 _Originally posted by @rolandgeider in [#1847](https://github.com/wger-project/wger/issues/1847#issuecomment-2979927721)_

As stated above, this should be pretty doable.

betterrrrr:
Hi @rolandgeider,

As discussed in #1989, this PR updates the AuthProxyHeaderMiddleware to support CIDR subnet notation for trusted proxy IPs.

Changes:

Imported the built-in ipaddress module.

Updated the IP matching logic in process_request to check if the client_ip belongs to any network defined in AUTH_PROXY_TRUSTED_IPS.

Used strict=False in ip_network to ensure backward compatibility (it works perfectly with both single IPs like 192.168.1.1 and subnets like 10.0.0.0/8).

Added basic try-except blocks to handle invalid formats gracefully without crashing the app.

Added 3 unit tests in test_auth_proxy_middleware.py to cover subnet matching, out-of-subnet rejection, and invalid config handling.

Fixes #1989.

### DIFF
--- a/wger/core/middleware.py
+++ b/wger/core/middleware.py
@@ -15,6 +15,8 @@
 # Standard Library
 import logging
 
+import ipaddress  # Support IP and CIDR subnet parsing
+
 # Django
 from django.conf import settings
 from django.contrib.auth import (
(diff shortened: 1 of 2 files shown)


rolandgeider:
merged, thanks!

Issue #2237, opened 2026-03-02
Import serving size from OFF


rolandgeider:
We should extract the serving size from the OFF data and generate the appropriate per-ingredient units (or update them). This would make logging ingredients much easier, eee e.g. https://github.com/wger-project/react/issues/1174

TODO:
* update `wger.nutrition.extract_info.off.extract_info_from_off` so that the serving size is read, if available
* update or create the unit
* probably also needed: refactor the way the unit tables are done, this can be probably done in an easier way

lvdwijngaart:
Hi, we would also like to work on this one. We will start on this now so hopefully we can be assigned :)

rolandgeider:
it's yours :)

paigefletcher1282-code:
Any updates on this

PR #2261, opened 2026-03-22
Off serving size

lvdwijngaart:
# Proposed Changes

- Serving sizes are introduced and fetched from OFF. 
- In the ingredient overview, macro's can be calculated per serving size aswell (prior it was only per x grams)

## Related Issue(s)

- Closes issue #2237. 


rolandgeider:
merged, thanks!

PR #2275, opened 2026-04-08
Fix(manager) : Make weight and rep units mandatory #2203

### LINKED ISSUE #2203: Make weight and rep units mandatory
Currently both the weight and repetition units in `slot_entry` and `log` are nullable and we treat that case as using reps and kg as their units. Removing this would make the code much simpler:

* Migration that fills up all null values with their defaults
* Flutter and react code: remove special cases

pankaj-basnet:
- This PR addresses issue #2203 by enforcing mandatory units for weight and repetitions across the `manager` app. By removing nullability from these fields, we simplify the client-side logic (Flutter/React) and ensure consistent data integrity across the platform.
- Makes `weight_unit` and `repetition_unit` non-nullable in `SlotEntry` and `WorkoutLog`.

# Proposed Changes
- Removed null=True, blank=True from repetition_unit and weight_unit in models code files
- Removed allow_null=True from SetConfigDataSerializer unit fields

* **Migrations**:
    * **0024**: Data migration to backfill all existing `NULL` units to default values (ID 1).
    * **0025**: Manual schema migration to enforce `NOT NULL` constraints.

## Related Issue(s)

Fixes #2203. 

NOTE: Todo tasks for react/flutter remaining.


nisselarsson:
I really like not having to set weight and reps when I add a new routine as I might need to discover whats appropriate when I actually do the exercise.

Issue #2295, opened 2026-04-14
Allow filtering ingredients by nutriscore


rolandgeider:
Expand the ingredient autocompleter so that the user can filter by nutriscore.

The main problem in this issue is probably how to present this. We want the user to be able to run different filters like "only A", "better than C", etc. in a nice UI (perhaps using a modified slider?). The backend should already support this, but we might need to update the filtersets, I'm not sure right now.

This needs to be done for react (`IngredientAutcompleter`) and flutter (`IngredientTypeahead`).

NB: for react it's probably a good idea to refactor a bit the `searchIngredient` function and pass e.g. a filter object or similar.

NBB: probably a good idea to wait till #2286 is merged

kiranannadatha8:
Hi! I'm picking this up — starting with the backend piece.

Opened #2305 which extends `IngredientFilterSet` to accept `gt/gte/lt/lte` lookups on `nutriscore` (in addition to the existing `exact/in`), so the frontend autocompleters can issue range queries like "better than C" (`?nutriscore__lt=c`).

React (`IngredientAutocompleter`) and Flutter (`IngredientTypeahead`) follow-ups will come as separate PRs on their respective repos.

kiranannadatha8:
Frontend (React) side is now up as a PR: wger-project/react#1237 — it adds the switch + MUI slider to the ingredient autocompleter and wires it to the new `nutriscore__lte` lookup.

**Status across repos:**
- Backend (Django): #2305 — adds `gt/gte/lt/lte` range lookups to the nutriscore filter
- Frontend (React): wger-project/react#1237 — adds the UI + sends `nutriscore__lte` on search
- Frontend (Flutter): wger-project/flutter#1179 — ports the same UI (switch + A–E discrete slider) to the mobile client and threads `nutriscoreMax` through `NutritionPlansProvider.searchIngredient`

Happy to rebase / split / adjust once reviewers take a look.

PR #2305, opened 2026-04-19
feat(nutrition): allow range lookups on nutriscore ingredient filter

kiranannadatha8:
## Summary
- Extend `IngredientFilterSet` so `nutriscore` accepts `gt`, `gte`, `lt`, `lte` lookups in addition to the existing `exact` and `in`. This enables range queries like "better than C" (`?nutriscore__lt=c`) that the React `IngredientAutocompleter` and Flutter `IngredientTypeahead` will consume.
- Since `nutriscore` is stored as a single lowercase character in the A–E range, lexicographic comparison on the `CharField` matches the grade ordering (A is best). `NULL` grades are excluded from range lookups by the Django ORM default, which matches expectations.
- Backend-only change — UI follow-up in `wger-project/react` and `wger-project/flutter` will be submitted separately.

Refs #2295 (blocker #2286 is merged).

## Test plan
- [x] New unit tests in `wger/nutrition/tests/test_search_api.py` covering `exact`, `in`, `lt`, `lte`, `gt`, `gte` on `nutriscore`
- [x] `uv run ./manage.py test wger.nutrition.tests.test_search_api` — 13 tests pass
- [x] `uv run ./manage.py test wger.nutrition` — 251 tests pass, no regressions
- [ ] Manual spot check via API: `GET /api/v2/ingredient/?nutriscore__lt=c` returns only A/B grades

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,5 @@
 # Changelog for the next release
 
+* The ingredient API filter now supports range lookups on `nutriscore`
+  (`gt`, `gte`, `lt`, `lte`) in addition to `exact` and `in`, enabling queries
+  like "better than C" (#2295).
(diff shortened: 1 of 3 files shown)


rolandgeider:
merged, thanks

PR #2316, opened 2026-04-26
Use "dumbbells" in Biceps Curls With Dumbbell

baijum:
# Proposed Changes

Fix description typo

- [ ] Code has been formatted to avoid unnecessary diffs (`ruff format && isort .`)


Abarbesgaard (approved):
see no issue here

rolandgeider:
Hi and thanks for the PR. Those fixtures are generated once in a while, typically for a release, so they would be overwritten the next time we do that. The "source of truth" is the upstream db, if you have an account on wger.de you can edit the descriptions directly there

PR #2327, opened 2026-05-06
Fix/preferences email validation

### LINKED ISSUE #2326: Preferences page shows no error message when an invalid email is entered
### Priority/Impact

Low (minor inconvenience)

### Description

Hello, I have been navigating the wger application adding workouts and realised I had entered my email incorrectly. When I tried to change it in my preferences and clicked Save, nothing was happening. I then realised there was no validation to tell you whether your email is valid or not when trying to update it — the page just reloads silently with no feedback.

**To Reproduce**

1. Register a new account/log in
2. Navigate to My Preferences (click your username in the top right)
3. Enter an invalid email address 
4. Click Save

**Expected behaviour**
An error message should appear informing the user that the email address entered is invalid.

**Actual behaviour**
The page reloads with no error message or success message, so a user would not know if there changes have been saved.

I also noticed this applies to all required fields, if you clear a required field and click Save, the same silent failure occurs with no error message to indicate what went wrong.

I am happy to work on a fix for this if it would be welcomed.

### Server version

wger 2.6.0-alpha2

### Mobile app version

_No response_

BigJimmy333:
# Proposed Changes

- Fixed the preferences page not showing any error message when a user enters an invalid email, a duplicate email, or leaves a required field empty as the form would do nothing, leaving the user with no feedback

- Fixed error messages not displaying correctly across the app due to an incorrect CSS class in the base templates

- Added novalidate to the preferences form so the browser no longer silently blocks invalid input before it reaches Django

## Related Issue(s)

Fixes #2326

### DIFF
--- a/wger/core/forms.py
+++ b/wger/core/forms.py
@@ -171,6 +171,10 @@ class UserPreferencesForm(forms.ModelForm):
 
         self.helper = FormHelper()
         self.helper.form_class = 'wger-form'
+        # Disable browser-side HTML5 validation so invalid emails and empty
+        # required fields submit through to Django, which surfaces a proper
+        # error message inline instead of the browser's own popover.
+        self.helper.attrs = {'novalidate': True}
         self.helper.layout = Layout(
             Fieldset(
                 _('Personal data'),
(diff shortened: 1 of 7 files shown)


rolandgeider:
good catch! The form handling in the prefs is definitely something that should be refactored in the future

PR #2338, opened 2026-05-12
Feature/dynamic measurements

### LINKED ISSUE #1560: Dynamic measurements
Add the possibility to create dynamic measurements that are calcualted based on other entries. For example, a 1RM calculated value for squats from the logged values or the BMI.

How exactly this would be configured needs to be discussed first, config file, small python module, etc. since we might want some logic there (in the 1RM example, we might only want to take into account sets with 5 reps or less, or just take the entry with the highest weight for a specific day)

dylans-dev:
# Proposed Changes

- Add support for dynamic measurement categories. 
- The /api/v2/measurement/ endpoint now intercepts requests for dynamic categories (such as BMI) and auto-calculates the results on the fly.
- A new endpoint at /api/v2/measurement-category/dynamic-types/ has been added to expose the available dynamic options.
- An enum is used for the new dynamic options with BMI currently and can be extended for more options and calculations.
- Utils package added with bmi module for BMI calculation.

## Related Issue(s)

- Related to this: https://github.com/wger-project/wger/issues/1560
- For required front-end React changes: https://github.com/wger-project/react/pull/1248


wger/measurements/models/category.py
```diff
@@ -41,6 +41,15 @@ class Meta:
         max_length=30,
     )
 
+    class DynamicType(models.TextChoices):
+        NONE = ('NONE',)
+        BMI = ('BMI',)
+        # SQUAT_1RM = 'SQUAT_1RM', can be added in future
+
+    dynamic_type = models.CharField(
+        max_length=20, choices=DynamicType.choices, default=DynamicType.NONE, db_index=True
```
rolandgeider, line 50: I would also add a json field here to configure the calculations, similarly to the one in the trophies. Then for a 1RM we can save the exercise ID (or other config-like things like using only up to 5 reps to calculate this) and don't need to add one function for each exercise

wger/measurements/migrations/0005_remove_category_is_dynamic_category_dynamic_type.py
rolandgeider, line 1: can you squash these two migrations?

wger/measurements/utils/bmi.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Workout Manager.  If not, see <http://www.gnu.org/licenses/>.
+
+
+def calculate_bmi(user, category_id):
+    # wger
+    from wger.weight.models import WeightEntry
+
+    profile = getattr(user, 'userprofile', None)
```
rolandgeider: users always have a profile

wger/measurements/utils/bmi.py
```diff
@@ -0,0 +1,41 @@
...
+    # height_sq will be a float
+    height_sq = (profile.height / 100) ** 2
+
+    weights = WeightEntry.objects.filter(user=user).order_by('date')
+
+    return [
+        {
+            'id': w.id,  # Use integer IDs so the frontend state manager doesn't complain
+            'category': int(category_id),  # Link it to the requested category
+            'date': w.date.isoformat() if hasattr(w.date, 'isoformat') else w.date,
```
rolandgeider: and entries always have a date :)

wger/measurements/api/views.py
```diff
@@ -101,3 +123,29 @@ def get_queryset(self):
...
+                # look up the category and check its enum value
+                category = Category.objects.get(id=category_id, user=request.user)
+
+                if category.dynamic_type != Category.DynamicType.NONE:
+                    calc_func = DYNAMIC_REGISTRY.get(category.dynamic_type)
+
+                    if calc_func:
+                        # execute the math function and wrap it in the DRF pagination envelope
+                        data = calc_func(request.user, category_id)
+                        return Response(
```
rolandgeider: I wonder if we could use DRF's pagination and serializer infrastructure instead of just returning all this in one go

wger/measurements/models/category.py
dylans-dev: Yeah will do, makes more sense for scalability.

dylans-dev:
Hey, I've made the requested changes regarding adding dynamic calc params, pagination and cleaned up migrations.
I also want to add a validator shortly with an example of what params should be included for instance with a 1 rep max calculation, since I'm running short on time to implement it in full

dylans-dev:
Hey, sorry I forgot to review my tests after I made those requested changes, the tests pass on my end now after that latest commit.

rolandgeider:
> Hey, sorry I forgot to review my tests 

don't worry, that's why we have 'em! 😄

dylans-dev:
> > Hey, sorry I forgot to review my tests
> 
> don't worry, that's why we have 'em! 😄

Haha it always happens to me, I'll work on fixing the front end now too.

rolandgeider:
FYI, since we're getting close to finishing a big release (see e.g. https://github.com/wger-project/flutter/pull/1183), I'm moving this to the next version 2.7 since the dynamic measurements come with some changes to the backend and frontend and I'd be more comfortable with some time

PR #2343, opened 2026-05-17
Fix: Handle missing product_name in Open Food Facts import

Ewin7:
# Proposed Changes

This PR relaxes the strict `product_name` check during the Open Food Facts (OFF) import to prevent valid products from being rejected.

### Why?
Currently, `product_name` is listed in `OFF_REQUIRED_TOP_LEVEL`. However, Open Food Facts is sometimes inconsistent. For some products, the  `product_name` key is missing, and the API only returns localized versions like `product_name_en` or `product_name_de`.

### How I fixed it:
1. Removed `'product_name'` from `OFF_REQUIRED_TOP_LEVEL`.
2. Implemented a fallback logic in `extract_info_from_off`:
- First, it tries the generic `product_name` or `product_name_en`.
- If not found, it iterates through the keys to find any localized `product_name_*`.

To test you can use the barcode: `4260648131900`.

### DIFF
--- a/wger/nutrition/extract_info/off.py
+++ b/wger/nutrition/extract_info/off.py
@@ -28,7 +28,6 @@ from wger.utils.constants import ODBL_LICENSE_ID
 
 
 OFF_REQUIRED_TOP_LEVEL = [
-    'product_name',
     'code',
     'nutriments',
 ]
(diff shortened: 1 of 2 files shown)


rolandgeider:
yes, good idea. We will probably need to do something similar if we ever start extracting the categories from the products

Issue #2145, opened 2025-12-14
Username suggestion system


Mogillaakhil:
### Use case

while new user is registering in the app , if there was a suggestion system for user name it will be better.
the username which are suggested by the system must be unique from other users .

![Image](https://github.com/user-attachments/assets/e7b88c2c-ffc9-4d0a-a686-800e6369a00a)

### Proposal

it will be easy for new user

jshudbrd:
Hey! I would like to work on this issue.

PR #2344, opened 2026-05-18
Feature/username suggestion system

jshudbrd:
# Proposed Changes

- Adds a shared `generate_username_suggestions` utility function in `wger/utils/username.py` 
  used by both the serializer and web registration form validator.
- Replaces the default DRF `UniqueValidator` on the username field with a custom 
  `validate_username` method that generates and returns available username suggestions
  when the chosen username is already taken.

## Related Issue(s)

- related to https://github.com/wger-project/wger/issues/2145


jshudbrd:
As a side note, a more complete solution would probably involve creating a registration component in the React repo that calls a username suggestions endpoint as the user types, displaying suggestions as clickable elements before the form is submitted. This would provide a better UX than the current approach of returning suggestions in an error response after a failed submit.

wger/core/api/serializers.py
```diff
@@ -164,14 +165,23 @@ class UserRegistrationSerializer(serializers.ModelSerializer):
     email = serializers.EmailField(required=False)
     username = serializers.CharField(
         required=True,
-        validators=[UniqueValidator(queryset=User.objects.all())],
+        # validators=[UniqueValidator(queryset=User.objects.all())],
```
rolandgeider: you can just remove this altogether :)

wger/core/api/serializers.py
```diff
@@ -164,14 +165,23 @@ class UserRegistrationSerializer(serializers.ModelSerializer):
...
     class Meta:
         model = User
         fields = ('username', 'email', 'password')
 
+    def validate_username(self, value):
+        if User.objects.filter(username__iexact=value).exists():
+            suggestions = generate_username_suggestions(value)
+            suggestions_string = ', '.join(suggestions)
+            raise serializers.ValidationError(
+                f'A user with this username already exists. \nSuggestions: {suggestions_string}'
```
rolandgeider, line 179: does the `\n` survive the json parsing trip? Better just remove it

wger/core/forms.py
```diff
@@ -318,6 +329,16 @@ class RegistrationFormNoCaptcha(UserCreationForm, UserEmailForm):
     Registration form without CAPTCHA field
     """
 
+    def clean_username(self):
```
rolandgeider, line 327: this is duplicated, perhaps extract it to something like `core/validators.py`?

wger/core/forms.py
```diff
@@ -318,6 +329,16 @@ class RegistrationFormNoCaptcha(UserCreationForm, UserEmailForm):
...
+        username = self.cleaned_data.get('username')
+        if User.objects.filter(username__iexact=username).exists():
```
rolandgeider: I think the django's validator is case sensitive, while this is probably the more usual approach, I would need to check first if we would get collisions, better just use `__exact` to be on the safe side

wger/utils/username.py
```diff
@@ -0,0 +1,40 @@
...
+def generate_username_suggestions(username: str, count: int = 3):
+    """
+    Given a username, generates a list of available unique usernames.
+    """
+    text = re.sub(r'[^\w]', '', username)
+
+    candidates = [
+        f'{text}{random.randint(1, 99)}',
+        f'{text}{random.randint(100, 999)}',
+        f'{text[0]}{text}{random.randint(1, 10)}',
```
rolandgeider, line 32: note that this will blow up if the username is made up completely of symbols or such as text will be an empty string. I'd say we just set text to "user" if that's the case

rolandgeider:
yes, a "real" feedback is definitely the best solution, but this should already be an improvement

wger/core/api/serializers.py
jshudbrd: This was just for formatting on the flutter app. I can remove it.

rolandgeider: that was my question, wasn't sure it would survive, we can leave it

wger/core/api/serializers.py
jshudbrd: Sure thing!

wger/core/forms.py
jshudbrd: Good call, I can extract to ```wger/core/validators.py``` and both form classes can then call a validate function from there

wger/utils/username.py
jshudbrd: Good catch, I can default the text to 'user'. Also need to write a test to cover this scenario

rolandgeider:
Merged, thanks

Issue #2340, opened 2026-05-14
Improve ingredient search


rolandgeider:
At the moment, the ingredient search can be a bit... problematic when you want to find a specific product.

Some ideas:

* add an exact match, similar to what the exercises already support
* push ingredients already used by the user to the top (should probably also be user-controllable and would need some discussion first)
* control the number of results (can we do something like infinite scroll when we reach the bottom?)
* ~~better search results~~
* ...

Ewin7:
A small improvement to the search would be to automatically look up manually entered barcode numbers in the Open Food Facts database. Right now, it only searches on camera scan. I can implement this, and maybe work on further improvements in the future.

anjakDev:
What do think about adding `common_name` and `brand` to the existing trigram search on the name? It would make use of the `common_name` value, and searching for the brand might sometimes be useful too. Neither of these fields should be its separate search bar imo, but as an 'implicit' inclusion to the existing search field, it might make the hits better. I'd be happy to implement that, but since it would need a migration on the indexes of `ingredients.py` for the new GIN indexes on `common_name` and `brand`, I wanted to give a heads up and ask for your opinion.

anjakDev:
As for the one improvement you (@rolandgeider ) already mentioned, I'd be a big fan of sorting up the items that a user has previously logged. I think we can easily make use of the `LogItems` entries for that and from a UX standpoint, it makes a lot of sense to show historically logged entries. It's very common to repeatedly eat and therefore log the same items. Maybe at some point we could even add an info like 'last logged on [insert-date]' in the details view or maybe even the search results list...

rolandgeider:
oh yes, we definitely don't want a second search box for the brand.

We can add new indices for those, but we would need to do some tests with the full db and see if we really want to always search all three fields. While I guess users expect the search to be a bit magic, but from personal experience it's very frustrating when the returned results have (seemingly) nothing to do with what you entered. But I guess we can also just add those to the expanded search options and select which fields should be searched.

+1 for the 'last logged on [insert-date]', but that will probably need more changes to make it work

MikaelSiidorow:
Following up on the concern about returning seemingly unrelated results, I looked into the current whole-name trigram search. Short terms can be excluded even when they appear literally in the ingredient name:

- `chicken` does not find `Chicken Wing, Marinated, Raw` (**0.296**).
- `bread` does not find `Rye Bread, Rye Mixed Grain Bread` (**0.273**).
[...]

PR #2351, opened 2026-05-21
feat: trigger Open Food Facts lookup on manual barcode text search

Ewin7:
# Proposed Changes

- Introduces a small improvement to the ingredient search, addressing my comment in #2340.
- Currently OFF lookups are only triggered when scanning a barcode via the camera. If a user manually enters a barcode into the search field, the system defaults to text search.
- This PR tries a barcode lookup first when the input looks like a barcode, and otherwise falls back to standard full-text search.

## Changes
- Updated `search_name_fulltext` in `nutrition/api/filtersets.py`. If the input is numeric and matches standard retail barcode lengths (8, 12, 13, 14 digits) the query is passed to `search_barcode`.
- Unit Testing: Added tests in `nutrition/tests/test_search_api.py`.

## Related Issue(s)
- Addresses my comment in #2340.

### DIFF
--- a/wger/nutrition/tests/test_search_api.py
+++ b/wger/nutrition/tests/test_search_api.py
@@ -22,6 +22,7 @@ from rest_framework import status
 # wger
 from wger.core.tests.api_base_test import ApiBaseTestCase
 from wger.core.tests.base_testcase import BaseTestCase
+from wger.nutrition.models import Ingredient
 
 
 class SearchIngredientApiTestCase(BaseTestCase, ApiBaseTestCase):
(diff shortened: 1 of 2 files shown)


wger/nutrition/tests/test_search_api.py
```diff
@@ -97,6 +100,38 @@ def test_search_all_languages(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data['count'], 7)
 
+    @patch('wger.nutrition.models.Ingredient.fetch_ingredient_from_off')
+    def test_search_name_with_valid_barcode(self, mock_fetch):
+        """
+        Searching for an 8-14 digit number in the name field should trigger the OFF API lookup
+        """
+        self.client.get(self.url + '?name__search=1300000000000')
+        mock_fetch.assert_called_once_with('1300000000000')
```
rolandgeider: we could simplify the tests here, just mock `search_barcode` and only test that the correct parameters are passed, that search_barcode itself works is already tested somewhere else (plus one that checks that the search directly returns if the search found an ingredient)

Ewin7: I updated the tests to mock `search_barcode`. I also updated the first test to verify the correct arguments are passed, and added a test to check if a barcode match is returned in the search response.

rolandgeider:
merged, thanks!

PR #2355, opened 2026-05-25
fix: barcode fetch result was never returned

anjakDev:
# Proposed Changes
### Before
When a client sent `GET /api/v2/ingredient/?code=<barcode>` for a barcode not yet in the local database, the method correctly called `Ingredient.fetch_ingredient_from_off(value)`, which fetched the product from Open Food Facts and saved it as a new Ingredient row. However, the return value of that call was discarded. The `queryset` variable, which was already bound to `queryset.filter(code=value)` with zero results, was returned unchanged. The API response therefore always contained an empty result set for any barcode that wasn't already local, even when the OFF fetch succeeded and the ingredient now existed in the database.

 ### After
The return value of `fetch_ingredient_from_off` is captured. If it is not `None` (i.e. the ingredient was successfully created), the result `queryset` is reassigned to `queryset.filter(pk=ingredient.pk)`, which resolves to the newly created row. The API response now correctly returns that ingredient. Filtering by `pk` rather than re-applying `code=value` also makes the fix robust against any barcode normalisation that OFF may apply before saving (e.g. stripping leading zeros), where a re-filter on the original code string could still return empty.

## Related Issue(s)

Tries to improve _ingredient search_ as per the general points described in #2340 

### DIFF
--- a/wger/nutrition/api/filtersets.py
+++ b/wger/nutrition/api/filtersets.py
@@ -61,12 +61,14 @@ class IngredientFilterSet(filters.FilterSet):
         if not value:
             return queryset
 
-        queryset = queryset.filter(code=value)
-        if queryset.count() == 0:
+        result = queryset.filter(code=value)
+        if not result.exists():
             logger.debug('barcode not found locally, trying to fetch ingredient from OFF')
-            Ingredient.fetch_ingredient_from_off(value)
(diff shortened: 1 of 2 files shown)


rolandgeider:
thanks!

PR #2356, opened 2026-05-25
fix: unknown language code is ignored instead of defaulted to english

anjakDev:
# Proposed Changes
When filtering in the _Ingredient Search_ for languages that are unknown to the `wger` backend, the existing comment (`line 99`) suggests any unknown code is simply ignored, however I noticed it ended up being defaulted to _English_ instead (which would add _English_ to the list of languages for the results if it wasn't already in the list). 

If that is intended behavior, feel free to close without merging. 

If, however, it is not intended, I propose this slight rewrite to make sure unknown codes do NOT default to `en` and would simply not be considered a valid filter for search results. The existing test (`test_search_unknown_language_codes(self)`) did not catch this behavior because it had `en` as part of the set already, and therefore the `en` data items were included anyway.

## Related Issue(s)

Is a further attempt at improving the _ingredient search_ as mentioned in #2340 . 

I am trying to add some smaller fixes before trying myself in bigger search improvements, just to get a better understanding of the code base first. 

### DIFF
--- a/wger/nutrition/api/filtersets.py
+++ b/wger/nutrition/api/filtersets.py
@@ -23,6 +23,7 @@ from django.contrib.postgres.search import TrigramSimilarity
 from django_filters import rest_framework as filters
 
 # wger
+from wger.core.models import Language
 from wger.nutrition.models import (
     Ingredient,
     LogItem,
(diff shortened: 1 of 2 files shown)


rolandgeider:
thanks for the PR, no you are right that we shouldn't default to English in this case

PR #2397, opened 2026-06-18
Fix: Allow overnight workouts and prevent session splitting (#2379)

### LINKED ISSUE #2379: Training over night breaks routine into two
### Priority/Impact

None

### Description

Heho
My training is often at night.
If I start at 23:00 and end at 01:00 the training get split into two trainings.
I also cant create a Log because the "end time can not be before the start time".

### Server version

2.6.0a2+gitea78447

### Mobile app version

1.11.0

prkhrexists:
This PR resolves the overnight workout bug where sessions crossing midnight trigger validation errors and get split into two separate logs.

Changes Made:

session.py: Updated WorkoutSession.clean(). Valid overnight sessions (e.g., 23:00 → 01:00) now pass validation. To prevent absurd spans (e.g., AM/PM typos), overnight durations are calculated and capped at a maximum of 12 hours. Removed a dead import to maintain linting standards.

log.py: Updated WorkoutLog.save(). Removed reliance on wall-clock times. When a new log is created, the logic now checks the latest session state for that user/routine. It only shifts the new log back by one day if the latest session is from yesterday, explicitly overnight (start > end), or still active (start set, end empty).

Tests: Added regression tests to cover the anchored overnight session cases.

Fixes #2379


rolandgeider:
Thanks for the PR! I still need to think if this is the right approach, or if we shouldn't refactor how we handle logs and sessions (e.g. changing the start and end fields to proper datetimes would make the "end not before start" also work across day boundaries. Also all the "magic" we have when saving logs).

In any case I'm adding the issue to the next release so it doesn't get lost

prkhrexists:
I would love to contribute to the refactoring approach you mentioned. Could you elaborate a bit more on the proposed solution so I can get started?

rolandgeider:
@prkhrexists alright, I have added some ideas to the original issue. See if they make sense to you

PR #2415, opened 2026-06-27
Require POST for password reset and user (de)activation

### LINKED ISSUE #2380: GET-based password reset and account deactivation can be triggered without CSRF protection
# Bug Report Draft

Issue form: `.github/ISSUE_TEMPLATE/1_bug.yml`

## Priority/Impact

High (critical issue, blocks workflow)

## Description

Two account-management actions perform security-sensitive state changes through GET requests:

1. `reset_user_password()` at `wger/gym/views/gym.py:243-264` resets a selected user's password without requiring POST.
2. `UserDeactivateView.get_redirect_url()` at `wger/core/views/user.py:402-406` deactivates a selected user from a `RedirectView`, so the mutation is triggered by GET.

Because Django CSRF protection does not apply to GET requests, an attacker can cause an authenticated gym manager/admin/trainer with sufficient permissions to load a crafted URL and unintentionally reset a user's password or deactivate a user account.

### Steps to reproduce

1. Log in as a user who has permission to manage a target gym user.
2. Visit `/gym/user/<user_pk>/reset-user-password`.
3. Observe that the target user's password is changed during a GET request.
4. Visit `/user/<pk>/deactivate`.
5. Observe that the target user's `is_active` flag is set to `False` during a GET request.

### Expected behavior

Password reset and account deactivation should require a POST request protected by CSRF validation and should involve an explicit confirmation step.

### Actual behavior

Both operations are reachable as GET-driven state changes. A third-party page can trigger the browser to request those URLs with the operator's session cookie.

### Security impact

The target user can be locked out of their account without the manager's deliberate consent. For password reset, the user's old password is invalidated and a new generated password is rendered in the manager's browser response.

### Suggested fix

Use POST-only views for these actions, keep Django CSRF protection enabled, and render confirmation forms before applying the mutation. `UserActivateView` should be reviewed at the same time because it follows the same `RedirectView` state-change pattern.

## Server version

Current `master` at `ea7844731f7e3fe5865c327c1f93f5f4a76dde75`; `wger/version.py` reports `2.6.0-alpha2`.

## Mobile app version

Not applicable.

amitmishra11:
## Why

Three views performed sensitive state changes as a side effect of a GET request, with no confirmation step:

- `reset_user_password()` (`wger/gym/views/gym.py`) resets a user's password to a random value.
- `UserDeactivateView.get_redirect_url()` and `UserActivateView.get_redirect_url()` (`wger/core/views/user.py`) flip `is_active` on a user, both `RedirectView`s that mutated state while computing the redirect target for *any* request, including GET.

Django's CSRF protection only applies to unsafe HTTP methods (GET is exempt by design), so a third-party page could embed a link/image pointing at these URLs and, if loaded by an authenticated gym manager/trainer's browser (which sends the session cookie automatically), trigger a password reset or account deactivation without the manager's consent — a classic CSRF-via-state-changing-GET.

This codebase already has precedent for fixing this exact class of bug: `trainer_login` in `wger/core/views/user.py` was previously converted to `@require_POST`, with the comment "rebinding the session is a state change and must go through Django's CSRF protection, which only applies to unsafe HTTP methods." This PR applies the same principle to the three remaining GET-mutates endpoints.

## Fix

- `reset_user_password`: GET now renders a confirmation form (reusing the existing `delete.html` template + crispy-forms pattern already used elsewhere for delete confirmations); only POST performs the actual reset.
- `UserDeactivateView` / `UserActivateView`: converted from `RedirectView` (which mutated unconditionally) to a small shared `UserActivationConfirmMixin` with explicit `get()` (renders the same confirmation pattern) and `post()` (performs the actual `is_active` change). All existing permission checks in `dispatch()` are unchanged.
- Updated `wger/core/templates/user/overview.html`: the "Reset user password" dropdown item now uses the existing `modal_link` tag (the same htmx-into-modal confirmation mechanism already used for "Deactivate/Activate user" and "Delete"), instead of a bespoke modal + inline JS that did `window.location.href = url` (a GET navigation). This removes the one-off modal/script block entirely — no template/JS changes were needed for deactivate/activate since they already used `modal_link`; only their view-side GET handler needed to stop mutating.

## Testing

- Updated the existing `activate`/`deactivate` test helpers and `test_trainer_can_still_deactivate_regular_member` in `wger/core/tests/test_user.py` to POST (their "legitimate flow" assertions were unintentionally still relying on GET mutating state).
- Added `test_activate_get_does_not_mutate` / `test_deactivate_get_does_not_mutate` (regression: GET returns 200 and leaves `is_active` unchanged).
- Added `test_reset_password_get_does_not_mutate` / `test_reset_password_post_mutates` in `wger/gym/tests/test_user.py` (the existing `test_reset_password_allowed_same_gym` only checked the status code, never that GET actually left the password alone, or that POST does perform the reset).
- Ran the full `wger.core` and `wger.gym` test suites (511 tests) — all pass.
- Ran `ruff check` and `isort --check-only` on all changed files — clean.

I'm developing on Windows and used a portable CPython 3.12 build plus a local SQLite-backed run with `DJANGO_SETTINGS_MODULE=settings.ci` to execute the real test suite (rather than guessing at behavior), so this was tested against the actual fixtures and templates, not just read statically.

Fixes #2380.

### DIFF
--- a/wger/core/templates/confirm.html
+++ b/wger/core/templates/confirm.html
@@ -1,0 +1,7 @@
+{% load i18n crispy_forms_tags %}
+
+<h4>{{ title }}</h4>
+{% if confirm_message %}
+    <p>{{ confirm_message }}</p>
+{% endif %}
+{% crispy form %}
--- a/wger/core/templates/user/overview.html
+++ b/wger/core/templates/user/overview.html
@@ -241,13 +241,9 @@
                 {% url 'core:user:activate' current_user.pk as url %}
                 {% modal_link url=url text=text css_class='dropdown-item' %}
             {% endif %}
-        <a
-            data-url="{% url 'gym:gym:reset-user-password' current_user.pk %}"
-            data-bs-toggle="modal"
-            data-bs-target="#confirmation-modal"
-            class="dropdown-item">
-            {% translate "Reset user password" %}
-        </a>
+        {% translate 'Reset user password' as text %}
+        {% url 'gym:gym:reset-user-password' current_user.pk as url %}
+        {% modal_link url=url text=text css_class='dropdown-item' %}
         {% if perms.gym.manage_gym or perms.gym.manage_gyms %}
         {% if current_user.userprofile.gym %}
             {% translate 'Roles' as text %}
@@ -261,38 +257,6 @@
         {% modal_link url=url text=text css_class='dropdown-item' %}
     </div>
 </div>
-
-<div id="confirmation-modal" class="modal fade" tabindex="-1" role="dialog">
-    <div class="modal-dialog">
-        <div class="modal-content">
-            <div class="modal-header">
-                <h4 class="modal-title">{% translate 'Password reset confirmation' %}</h4>
-                <button type="button" class="close" data-bs-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-            </div>
-            <div class="modal-body">
-                <p>{% translate 'Do you really want to change this user&#39;s password?' %}</p>
-            </div>
-            <div class="modal-footer">
-                <button type="button" class="btn btn-danger" data-bs-dismiss="modal">{% translate 'No' %}</button>
-                <button type="button" class="btn btn-success">{% translate 'Yes' %}</button>
-            </div>
-        </div>
-    </div>
-</div>
-
-<script>
-    $(document).ready(function() {
-        $('#confirmation-modal').on('show.bs.modal', function (event) {
-            var button = $(event.relatedTarget);
-            var url = button.data('url');
-            var modal = $(this);
-
-            modal.find('button.btn-success').click(function() {
-                window.location.href = url;
-            });
-        });
-    });
-</script>
 
 {% endif %}
 <h4>{% translate "Details" %}</h4>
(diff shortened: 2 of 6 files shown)


munzzyy:
Verified this against current master (ea78447 era, now a few commits ahead) — the vulnerable GET-mutates code is exactly as described in #2380, and this PR closes it cleanly.

Checked a few things that weren't obvious from the diff alone:

- Permission ordering in `reset_user_password`: the new `if request.method != 'POST':` confirmation branch sits after all three existing permission checks (`is_authenticated`, `manage_gym(s)`, `is_same_gym`). An unauthorized user still gets `HttpResponseForbidden()` before ever reaching the confirmation form, so there's no new info leak about the target user to someone who shouldn't be able to act on them.
- Grepped the repo for any other `RedirectView`/`get_redirect_url` mutating-on-GET pattern outside the three views this PR touches — found none. So this covers the full blast radius of the bug class, including the `UserActivateView` follow-through the issue explicitly asked for, not just the two originally-listed views.
- The `activate`/`deactivate` test helper changes from `.get` to `.post` are correct fixes to tests that were unintentionally relying on the bug (they were asserting on the mutation that a GET request caused).

Only thing I'd flag, not a blocker: `UserActivationConfirmMixin.get_object_or_404` runs twice per request cycle now (once implicitly via `dispatch()`'s permission checks reading `edit_user`, again in `get`/`post`) — pre-existing pattern in this codebase though (same double-fetch shape as the old `get_redirect_url`), so not a regression this PR introduces.

This looks complete and ready to merge.

rolandgeider:
merged 🥳

PR #2428, opened 2026-07-11
Make the PowerSync auth token lifetime configurable

nico-fioretti:
## What

The JWT that the Flutter app requests to authenticate against the PowerSync service is created in `wger/core/api/powersync.py` with a **hard-coded 600 s (10 min) lifetime**:

```python
"exp": now + 600,
```

This PR makes that lifetime configurable via a new setting **`POWERSYNC_ACCESS_TOKEN_EXPIRES_IN`** (seconds), defaulting to the previous `600` so behaviour is unchanged out of the box. On the docker settings it is read from the environment, so self-hosters can raise it without patching the source.

## Why

On self-hosted instances the 10-minute lifetime forces the mobile client to refresh the sync token very frequently. When a refresh is missed (typically when the app resumes from the background after the token has already expired) the PowerSync stream can stall, and the only reliable recovery is clearing the app storage and logging in again. The server logs the expected `PSYNC_S2103 JWT has expired` on `/sync/stream` in those windows.

Being able to lengthen the token (e.g. to 1 h) sharply reduces how often the client sits with an expired token on resume, which reduces the frequency of these stalls, without any client-side change.

## Changes

- `settings/settings_global.py`: add `POWERSYNC_ACCESS_TOKEN_EXPIRES_IN = 600` default.
- `settings/main.py`: read `POWERSYNC_ACCESS_TOKEN_EXPIRES_IN` from the environment (docker deployments).
- `wger/core/api/powersync.py`: use the setting instead of the hard-coded `600`.
- `wger/core/tests/test_powersync.py`: add `PowerSyncTokenLifetimeTestCase` asserting the default and an override are honoured.

## Compatibility

Default is unchanged (600 s), so existing deployments behave exactly as before unless they opt in by setting the env var.

### DIFF
--- a/settings/main.py
+++ b/settings/main.py
@@ -109,6 +109,7 @@ JWT_PUBLIC_KEY = env.str('JWT_PUBLIC_KEY', '')
 JWT_PRIVATE_KEY = env.str('JWT_PRIVATE_KEY', '')
 POWERSYNC_URL_PATH = env.str('POWERSYNC_URL_PATH', 'ps')
 POWERSYNC_URL = env.str('POWERSYNC_URL', '')
+POWERSYNC_ACCESS_TOKEN_EXPIRES_IN = env.int('POWERSYNC_ACCESS_TOKEN_EXPIRES_IN', 600)
 if not DEBUG and any(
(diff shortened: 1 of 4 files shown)


rolandgeider:
Closing this PR as the token lifetime wasn't the culprit of the problems in the linked issue

PR #2436, opened 2026-07-17
fix(weight): store per-entry weight_unit to prevent silent data reinterpretation

### LINKED ISSUE #2377: Body weight entries are stored without a unit, so changing the weight-unit preference silently reinterprets all historical data
### Priority/Impact

Medium (affects some functionality)

### Description

Body weight entries (WeightEntry) are stored as a plain number with no associated unit. The kg/lb unit is only ever applied as a display label based on the user's current weight_unit preference — it is never recorded on the entry and the value is never converted. As a result, changing the preference retroactively changes the meaning of every existing entry.

  Steps to reproduce:
  1. Set weight-unit preference to lb.
  2. Add a body weight entry of 100 (intending 100 lb).
  3. Change the weight-unit preference to kg.
  4. The same entry now reads 100 kg.

 Expected: the entry recorded as 100 lb should continue to represent 100 lb (≈45.4 kg) regardless of the current preference.

 Actual: the stored number (100) is simply relabeled, so the entry's real-world value changes when the preference changes.

  Scope / root cause (for reference):
  - Backend: WeightEntry (wger/weight/models.py) has a weight DecimalField and no unit field. __str__ hardcodes kg, but UserProfile.bmi()
  (wger/core/models/profile.py) interprets the latest entry as the current preference unit — so even the backend is internally inconsistent about what unit the
  stored number is in.
  - React (wger-project/react): WeightForm, WeightEntry model/adapter, and the chart/overview do no unit handling — the raw number is saved and displayed;
  useMetric is never consulted.
  - Flutter (wger-project/flutter): same — the value is stored/displayed raw; the preference only switches the kg/lb label (weight_overview.dart,
  dashboard/widgets/weight.dart).

  By contrast, WorkoutLog does store a per-entry weight_unit (FK, required), so exercise logs are unaffected this issue is specific to body weight.

### Server version

2.6.0a2+gitea78447

### Mobile app version

1.11.0

mmustafasenoglu:
## Description

Fixes #2377

Body weight entries were stored as a plain number without a unit. The displayed weight relied entirely on the user's current weight_unit profile preference. When the preference was changed, all historical entries were silently reinterpreted under the new unit, making previously correct values wrong.

## Solution

Add a weight_unit CharField (kg/lb, default kg) to WeightEntry so each entry carries its own unit. Existing data is assumed to be in kg (the historical default).

## Changes

- wger/weight/models.py: Added weight_unit field, updated __str__ to use per-entry unit
- wger/weight/migrations/0006_weightentry_weight_unit.py: New migration adding the column
- wger/weight/api/serializers.py: Exposed weight_unit in API
- wger/weight/api/views.py: Default weight_unit from UserProfile on create
- wger/weight/api/filtersets.py: Added weight_unit filter
- wger/weight/helpers.py: CSV import attaches user's profile unit
- wger/weight/views.py: CSV export includes unit column
- wger/core/models/profile.py: user_bodyweight() sets unit from profile
- wger/core/templates/user/overview.html: Displays per-entry unit
- wger/weight/templates/import_csv_preview.html: Shows unit in preview
- wger/weight/fixtures/test-weight-data.json: Updated test data with explicit units
- wger/weight/tests/test_csv_export.py: Adjusted content length bound

## Backward Compatibility

- All existing entries default to kg via the migration
- API clients that don't send weight_unit get the model default (kg)
- The field is fully backward-compatible with existing API consumers


mmustafasenoglu:
Closing this PR - it has been inactive for over 2 weeks with no maintainer response.

PR #2486, opened 2026-08-15
Improve ingredient search relevance for long names

MikaelSiidorow:
# Proposed Changes

- Expand single-term searches with indexable literal substring matches while preserving the existing whole-name trigram candidates for typos.
- Rank exact names, prefixes, and whole-word matches first, then order the remaining candidates by whole-name trigram similarity.
- Bound short and punctuation-only searches to avoid unindexed catalog scans.
- Add PostgreSQL integration tests for long names, multi-word queries, ranking, partial words, typos, Unicode case handling, and short inputs.

This fixes cases where a single search term appears inside a longer ingredient name but falls below the whole-name trigram cutoff, such as `lunch` not finding `Lunch out (placeholder)`.

One- and two-character searches use exact matching. Punctuation-only searches return no results without querying PostgreSQL. Multi-word searches keep the existing trigram candidate path with the improved ranking.

This PR intentionally leaves the optional exact-match toggle out of scope. That requires separate API and client-side UX decisions and can be considered as a follow-up.

## Related Issue(s)

Related to #2340. This improves name-based search, but personalized ranking and catalog quality remain follow-up work.

Discussed on [Discord](https://discord.com/channels/754275926921052230/1537906071589621781).

## Testing

Run the focused tests in the `dev-postgres` environment:

```bash
docker compose exec web python3 manage.py test \
  wger.nutrition.tests.test_search_api.IngredientSearchRankingApiTestCase
```

All 12 added PostgreSQL integration tests pass. The existing PostgreSQL ingredient search tests also pass.

Query plans and count-plus-top-20 latency were checked against 3,057,932 ingredient names. The indexed paths use the existing trigram GIN index, so no migration is needed. Across 12 common queries, median overhead compared with `main` was about 5%, with higher cost for searches where the literal arm expands recall.

- [x] Tests for the changes have been added
- [x] Code has been formatted to avoid unnecessary diffs
- [x] A changelog entry has been added; no manual deployment steps are needed

### DIFF
--- a/wger/nutrition/api/filtersets.py
+++ b/wger/nutrition/api/filtersets.py
@@ -17,7 +17,10 @@
 import logging
 
 # Django
-from django.contrib.postgres.search import TrigramSimilarity
+from django.contrib.postgres.search import (
+    TrigramSimilarity,
+    TrigramStrictWordSimilarity,
+)
 from django.db.models import (
     Case,
     F,
@@ -39,6 +42,7 @@ from wger.nutrition.models import (
 from wger.utils.db import (
     PostgresILikeContains,
     PostgresILikeExact,
+    PostgresILikeStartsWith,
     is_postgres_db,
 )
 from wger.utils.language import load_language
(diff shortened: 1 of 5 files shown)


wger/nutrition/tests/test_search_api.py
```diff
@@ -251,3 +255,107 @@ def test_barcode_not_found_locally_off_returns_none(self, mock_fetch):
         mock_fetch.assert_called_once_with('0000000000000')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data['count'], 0)
+
+
+@skipUnless(connection.vendor == 'postgresql', 'PostgreSQL ingredient search only')
```
rolandgeider, line 263: just a FYI, at the moment CI only runs the tests with (in-memory) sqlite, but it might be a good idea to run these ones also on postgres, I think it's possible to tag tests and testcases. But that would be something for another issue

MikaelSiidorow: I could set up the GitHub Actions to run it vs a Postgres service as well, if that makes sense to expand here

wger/nutrition/api/filtersets.py
```diff
@@ -99,9 +102,12 @@ def search_name_fulltext(self, queryset, name, value):
...
-                queryset.filter(name__trigram_similar=value)
-                .annotate(similarity=TrigramSimilarity('name', value))
-                .order_by('-similarity', 'name')
+                queryset.filter(name__trigram_word_similar=value)
+                .annotate(
+                    word_similarity=TrigramWordSimilarity(value, 'name'),
+                    similarity=TrigramSimilarity('name', value),
+                )
+                .order_by('-word_similarity', '-similarity', 'name')
             )
```
MikaelSiidorow, line 155: I found some more search edge cases that this doesn't find, but probably should. I'll iterate still a bit more, and will create regression tests out of those

wger/nutrition/tests/test_search_api.py
rolandgeider: I'd do this after we merge this issue, in the end it's really just a nice-to-have. If you want to give it a try, take a look first at tagging/marking the tests, we really only want those pg-only tests in the run (at the moment the testsuite runs in under 2 minutes in CI and we would only slow things down)

wger/nutrition/api/filtersets.py
MikaelSiidorow: Finished iterating on this. The search now combines literal substring matches with the original trigram search, so it finds words inside longer names without making typo matches much stricter.

One- and two-character searches use indexed exact matching only to avoid expensive catalog scans. I added regression tests for both cases.

MikaelSiidorow:
Ready for another review. Since the initial draft, I have:

- preserved typo matches from the original trigram search
- bounded short and punctuation-only inputs
- refined exact, prefix, whole-word, and Unicode ranking
- limited literal expansion to single-term searches after full-catalog performance testing

wger/nutrition/tests/test_search_api.py
MikaelSiidorow: Makes sense. I kept CI unchanged and ran the focused tests locally against PostgreSQL. I'll leave tagged PostgreSQL CI for a separate follow-up

rolandgeider (approved):

Issue #2480, opened 2026-08-13
Support double progression: gate on the top of the rep range and require all sets to qualify


bentucker:
### Use case

Double progression is one of the most common progression schemes in
hypertrophy training: you work inside a rep range (say 3 sets of 8-12) at a
fixed weight, add reps from workout to workout, and only increase the weight
once you reach the top of the range - typically once ALL prescribed sets hit
the top. After the weight increase you drop back to the bottom of the range
and start climbing again.

The current requirements mechanism can't express this. The available rules
(weight, repetitions, rir, rest) gate against the displayed prescription of
the base config, which for a rep range is the bottom. With 3x8-12, a
WeightConfig gated on {"rules": ["repetitions"]} already fires when the
lifter logs 8 reps, i.e. immediately on re-entering the range, so the weight
ratchets up every workout and the "fill the range first" phase never
happens. There is no way to reference the top of the range (the
MaxRepetitionsConfig value), and no way to require that every prescribed set
qualifies rather than any single one.

### Proposal

Two additive extensions to the existing requirements JSON, no schema or
migration changes:

1. New rules max_repetitions and max_weight. They read the same log field as
   their base rule (repetitions, weight) but gate against the top of the
   prescribed range, i.e. the value walked for the maxrepetitions /
   maxweight config. A weight progression for classic double progression
   becomes: {"rules": ["max_repetitions"]}.

2. New optional boolean all_sets. When set, the gating iteration only
   qualifies if at least the prescribed number of sets was logged and every
   logged set meets all rules, instead of a single qualifying set being
   enough. Strict double progression is then:
   {"rules": ["max_repetitions"], "all_sets": true}.

This slots into the walk-based engine from #2465's follow-up rework quite
naturally: the thresholds map gains entries for the max fields (rounded like
their base field), and _requirements_met gets the all-sets branch with the
prescribed set count floored at 1 so empty or under-logged iterations never
qualify vacuously.

I have a working implementation on top of the current engine, including
validator support (unknown keys rejected, all_sets type-checked), about 20
regression tests for the new rules and flag, and a CHANGELOG entry. All of
wger.manager passes. Happy to open a PR if this direction sounds good to
you, and to adjust naming or semantics to whatever you prefer.

### Additional Context

- I have been running this in production (self-hosted) since June for a
  6-day routine where every exercise uses
  {"rules": ["max_repetitions"], "all_sets": true} together with per-gym
  weight rounding, so the semantics have seen a few hundred real workouts.
[...]

PR #2502, opened 2026-09-04
feat(manager): double progression via max_repetitions/max_weight rules and all_sets

bentucker:
# Proposed Changes

- Implement the server side of #2480: double progression support for
  progression requirements.
- New requirement rules max_repetitions and max_weight gate on the top of
  the prescribed range instead of the bottom. REQUIREMENT_RULES in consts.py
  maps each rule key to its log field and threshold field; the max_* rules
  read the same log field as their base rule but compare against the max
  config's displayed prescription, rounded like the base field.
- New optional all_sets flag on the requirements object: the gating
  iteration only qualifies when at least the prescribed number of sets was
  logged and every logged set meets all rules (single qualifying set
  remains the default). The prescribed count is floored at 1 so empty and
  under-logged iterations never qualify vacuously.
- Classic double progression is then
  {"rules": ["max_repetitions"], "all_sets": true}: work 8-12 reps at a
  fixed weight, add weight only once all sets reach 12.
- Validator support (unknown requirement keys rejected, all_sets
  type-checked), and unknown rule keys persisted past the serializer-only
  validator are treated as unmet instead of raising.
- About 20 regression tests for the new rules and the flag, plus a
  CHANGELOG entry. No migrations: requirements is an existing JSON field,
  and the change is backward compatible (existing requirements keep their
  meaning).

This has been running in production on my self-hosted instance since June
across several hundred workouts. The routine editor UI for the new keys is
deliberately out of scope here; aadishsanghvi expressed interest in picking
that part up in the issue.

## Related Issue(s)

Implements the server side of #2480 (leaving it open for the UI half).

- [x] Tests for the changes have been added (wger.manager passes, 319 tests)
- [x] ruff format and import sorting applied
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 
 ### Sync with Apple Health and Health Connect
 The mobile app can now import your body metrics from Apple Health (iOS) and Health Connect (Android). Once enabled in the settings, the data your smart scale, blood pressure monitor, smartwatch or what other health apps record is imported automatically and shows up alongside your manually entered entries. At the moment we support these:
+### Double progression support
+Progression requirements can now reference the top of the prescribed range via the new rules `max_repetitions` and `max_weight`, and the new `all_sets` flag requires every prescribed set to qualify. Together they enable classic double progression schemes ("work from 8 to 12 reps at a fixed weight, add weight only once all sets reach 12"), e.g. `{"rules": ["max_repetitions"], "all_sets": true}`.
+
 
 - body weight
 - body fat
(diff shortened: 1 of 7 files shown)


bentucker:
Note on the second commit: the Lint Code Base check flagged a pre-existing line in CHANGELOG.md (106) at 408 characters against markdownlint's 400 limit, introduced by the recent line-break removal on master. Wrapped it here so the PR goes green; happy to drop that commit if you'd rather fix it on master directly.

